PR #36, opened 2023-02-22
Symbolic jump

karmacoma-eth:
A simple experiment to see if we can lift the restriction that jump targets have to be concrete.

This works by:
- implementing an alternate entrypoint in `halmos.ctf`
- this sets up an execution and specifically looks for successful traces (e.g. end with `STOP` or `RETURN`)
- in sevm, mirroring the way `jumpi` branches are treated, but for potential jump targets (i.e. spawn new executions where the target is a valid jump destination)

Test with:
`python3 -m halmos.ctf 34381856FDFDFDFDFDFD5B00`

We expect the solver to return `msg_value = 6` to trigger a successful execution path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dynamic = ["version"]
 
 [project.scripts]
 halmos = "halmos.__main__:main"
+ctf = "halmos.ctf:main"
 
 [project.urls]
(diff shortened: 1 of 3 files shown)


daejunpark:
Merged via #48

PR #41, opened 2023-03-09
[feat] add SIGNEXTEND op code

zobront:
I'm new to Z3 so please double check this to make sure I'm not missing anything.

This PR adds the `SIGNEXTEND` op code (as well as a basic test to confirm it's working). The goal of the op code is to take in some bytes and return 32 bytes that represents the same number.

SIGNEXTEND uses two values from the stack: the byte length of the current value and the value itself.
- We don't need to use the byte length, so we simply pop that off.
- We pop off the value, convert it to an int with `int(str(value))` and run it through `con()` to turn it into a 256 bit `BitVecVal`. 
- We push the result back on the stack.

The simple test I wrote converts an `int8` to the opposite sign. This is implemented using `SIGNEXTEND` in Solidity, so the test fails with the previous halmos version. It quickly passes with these new changes.

Please let me know if you have any feedback, thanks.


zobront:
I just realized this approach assumes the value is a literal, which isn't the case. I'll implement on the bit level and should solve the problem.

daejunpark:
Thank you for your effort!  To make it work for symbolic values as well, you would need to use the SignExt bitvector operation: https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L4359

Also, I think that the byte length argument needs to be considered. For example, if the given byte length is smaller than the actual size of the value, what will happen? My assumption is that it will truncate the higher bytes and extend the sign, but I haven't verified this with the semantics.

zobront:
Thank you!

SignExt takes in `n` (the number of bits to add) and `a` (the original value).

So the natural thing to do would be pop the first value (the current byte length) and set `n = 256 - length`. That would expand it to 256 bits, which is what SIGNEXTEND does.

However, this was returning longer BitVecs than 256. After some digging, it looks like all the BitVecs popped off the stack are already 256 bits, regardless of the byte length.

Does that seem right? 

If so, then SIGNEXTEND has no computations to perform — all it needs to do is pop the length off the stack and leave the value.

```python
elif o.op[0] == 'SIGNEXTEND':
    ex.st.pop()
```

daejunpark:
You would first need to [`Extract`](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L4124) the last length bytes, and then `SignExt`.

zobront:
I'm not sure that's possible, because the length is a symbolic value and `Extract` requires `high` and `low` to be concrete.

daejunpark:
Ah I see. I think we can support only for the concrete length. In normal use cases, the length value should be concrete, I think.

daejunpark:
Also, note that if the length > 31, then it just returns the target value, based on yellow paper and other vm implementations, e.g., https://github.com/ethereum/py-evm/blob/0e5cae5ae82ea1a960a18db4c26833ec2512fdb2/eth/vm/logic/arithmetic.py#L167-L183

daejunpark:
There is actually a way to support symbolic length values. Since there are "only" 33 possible length values (0, 1, 2, ..., 31, and 32+), we can use a big nested [`If`](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L1377-L1397) term to represent each of the 33 possible outputs. This would look something like:
```
If(length == 0, SignExt(248, Extract(7, 0, value)),
If(length == 1, SignExt(240, Extract(15, 0, value)),
...
If(length ==31, SignExt(8, Extract(247, 0, value)), value)))....)
```

But, I'm concerned that this big `If` term may negatively impact performance. So, I think it would be better to include an optional flag that allows for the expansion of the 33 cases, while supporting only concrete length values by default.

Please feel free to implement only the concrete case in this PR. We can implement the symbolic case separately in another PR.

daejunpark:
Yes, you can use [`is_bv_value()`](https://github.com/Z3Prover/z3/blob/master/src/api/python/z3/z3.py#L3954-L3966) to check if the length is a concrete value. If the length is concrete, you can convert it to a Python integer by `int(str(w))`. Keep in mind that if `w` is symbolic, this conversion will result in a runtime exception.

zobront:
@daejunpark Thank you for all the help. I think this is all set now. Tests are running properly and raises `ValueError` exception if a symbolic value is passed.

src/halmos/sevm.py
```diff
@@ -1031,6 +1031,15 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps]:
                     w = ex.st.pop()
                     ex.st.push(LShR(ex.st.pop(), w)) # bvlshr
 
+                elif o.op[0] == 'SIGNEXTEND':
+                    w = ex.st.pop()
+                    if not is_bv_value(w): raise ValueError(w)
+
+                    w = int(str(w))
+                    if w > 31: continue
```
daejunpark, line 1039: The `continue` statement should not be used here, as there are remaining steps (i.e., increasing the PC counter and pushing the current node to the worklist) after this if block.

zobront: Fixed.

daejunpark (approved):

daejunpark:
Looks great, thank you so much for your effort, @zobront!

PR #56, opened 2023-04-18
add pc parsing to sevm

plotchy:
Currently `halmos --bytecode 5800` (PC STOP) will cause a parsing error as PC isn't an option in `sevm/run()`

This is a simple change, adding `EVM.PC` as a parsing option in `sevm.py`.

Curious how you'd like to test this and similar situations. Solidity isn't kind to using pc:
solidity_pc

I found this when I was compiling Huff and running halmos on it, though adding Huff support for testing may be overkill.

Would you be in favor of adding it in through `test_cli.py` or `test_sevm.py`?
```
def test_run_pc(args, options):
    hexcode = '5800'
    options['sym_jump'] = True
    exs = run_bytecode(hexcode, args, options)
    assert len(exs) == 1
    ex = exs[0]
    assert str(ex.pgm[ex.this][ex.pc].op[0]) == str(EVM.STOP)
```

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -1214,6 +1214,9 @@ class SEVM:
                 #   ex.st.push(f_chainid())
                     ex.st.push(con(1)) # for ethereum
 
+                elif opcode == EVM.PC:
+                    ex.st.push(con(ex.pc))
+
                 elif opcode == EVM.BLOCKHASH:
                     ex.st.push(f_blockhash(ex.st.pop()))
 


daejunpark:
Thank you for your contribution, @plotchy!  It's great!

To test this, you can add it to the existing `test_run` parameterized test. For example, you can add [the evm.codes test case](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code=%27~0x~1xJUMPDESTzesq2x~3xPUSH1%201%20l4x~6%20%7Bpreviouminstructionmtakem2%20bytes%7D%27~PCwwwlz%20%2F%2F%20Offx%5Cnw%20%20qt%20ms%20lzseq%01lmqwxz~_) as follows:

```
diff --git a/tests/test_sevm.py b/tests/test_sevm.py
index 5fd6000..9178552 100644
--- a/tests/test_sevm.py
+++ b/tests/test_sevm.py
@@ -70,6 +70,7 @@ def o(opcode):
     (BitVec('x', 256), '[]', 0, 'Extract(255, 248, x)'),
     (Concat(BitVecVal(int('6001', 16), 16), BitVec('x', 8), BitVecVal(0, 8)), '[1]', 2, 'x'),
     (Concat(BitVecVal(int('6101', 16), 16), BitVec('x', 8), BitVecVal(0, 8)), '[Concat(1, x)]', 3, '0'),
+    (BitVecVal(int('58585B5860015800', 16), 64), '[6, 1, 3, 1, 0]', 7, '0'),
 ])
 def test_run(hexcode, stack, pc, opcode, sevm, solver, storage):
     ex = mk_ex(hexcode, sevm, solver, storage, caller, this)
```

I can push the above test if you allow me to edit your PR, or you can add your own test too!

--

> I found this when I was compiling Huff and running halmos on it, though adding Huff support for testing may be overkill.

I think it would be great to add Huff support! It's on my todo list, but I haven't had the chance to work on it yet. Please let me know if you're interested in that, and we can discuss it further!

daejunpark (approved):

PR #60, opened 2023-05-05
feat: add invalid opcode

daejunpark:
add invalid opcode support by treating it as non-panic revert

### DIFF
--- a/tests/test/Invalid.t.sol
+++ b/tests/test/Invalid.t.sol
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.5.2;
+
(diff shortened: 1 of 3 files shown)


tests/test/Invalid.t.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.5.2;
+
+contract OldCompilerTest {
+
+    function testAssert(uint x) public {
```
karmacoma-eth, line 6: if we compile this test with a recent solc, it won't actually exercise the `invalid` opcode path, right? might be worth it to have a test where we hardcode an `invalid` opcode

PR #68, opened 2023-05-10
Add support for PUSH0

karmacoma-eth:
This effectively switches the supported EVM version to Shanghai, not sure if we want to explicitly support older versions (e.g. to model chains that don't support it).

This updates the evm_version in foundry.toml so that `forge test --match testPush0` returns the expected value.

### DIFF
--- a/tests/test/Opcode.t.sol
+++ b/tests/test/Opcode.t.sol
@@ -46,4 +46,27 @@ contract OpcodeTest is Test {
         assertEq(result1, result2);
     }
 
+    function testPush0() public {
+        // target bytecode is 0x365f5f37365ff3
+        //  36 CALLDATASIZE
+        //  5F PUSH0
+        //  5F PUSH0
+        //  37 CALLDATACOPY -> copies calldata at mem[0..calldatasize]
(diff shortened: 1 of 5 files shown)


tests/test/Opcode.t.sol
```diff
@@ -46,4 +46,27 @@ contract OpcodeTest is Test {
...
+        address target;
+        assembly {
+            mstore(0, deployCode)
+            target := create(/* value */ 0, /* offset */ 0x11, /* size */ 15)
+        }
+
+        (bool success, bytes memory result) = target.call(bytes("hello PUSH0"));
+        assertTrue(success);
+        assertEq(string(result), "hello PUSH0");
+    }
```
daejunpark, line 71: nice, you can also add a quick opcode test in test_opcode_simple() below

daejunpark (approved):

tests/test/Opcode.t.sol
karmacoma-eth: Oh nice I didn't realize we had pytest now 🤗

PR #72, opened 2023-05-12
Add support for vm.etch

karmacoma-eth:
Also:
- replaces mk_this and mk_caller with mk_addr(name)
- fixes an issue in wload when loading a single byte
- adds convenience method Opcode.__repr__, useful when printing
- adds convenience methods extract_bytes and extract_funsig
- adds convenience method decode_bytes, similar to decode_hex

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -138,14 +138,6 @@ def mk_calldata(abi: List, funname: str, funsig: str, arrlen: Dict, args: argpar
 def is_stop_or_return(opcode: Byte) -> bool:
     return is_bv_value(opcode) and opcode.as_long() in [EVM.STOP, EVM.RETURN]
 
-def decode_hex(hexcode: str) -> Tuple[List[Opcode], List[Any]]:
-    if hexcode.startswith('0x'):
-        hexcode = hexcode[2:]
-    if len(hexcode) % 2 != 0: raise ValueError(hexcode)
-    (ops, code) = decode(BitVecVal(int(hexcode, 16), (len(hexcode) // 2) * 8))
-    pgm = ops_to_pgm(ops)
-    return (pgm, code)
-
 def mk_callvalue() -> Word:
     return BitVec('msg_value', 256)
 
@@ -165,15 +157,9 @@ def mk_block() -> Block:
     block.chainid = con(1) # for ethereum
     return block
 
-def mk_caller(solver) -> Word:
-    caller = BitVec('msg_sender', 256)
-    solver.add(Extract(255, 160, caller) == BitVecVal(0, 96))
-    return caller
-
-def mk_this(solver) -> Word:
-    this = BitVec('this_address', 256)
-    solver.add(Extract(255, 160, this) == BitVecVal(0, 96))
-    return this
+# TODO: addresses are used as keys in the context and could be treated as 160-bit values
+def mk_addr(name: str) -> Word:
+    return ZeroExt(96, BitVec(name, 160))
 
 def mk_solver(args: argparse.Namespace):
     solver = SolverFor('QF_AUFBV') # quantifier-free bitvector + array theory; https://smtlib.cs.uiowa.edu/logics.shtml
@@ -190,8 +176,8 @@ def run_bytecode(hexcode: str, args: argparse.Namespace, options: Dict) -> List[
     balance = mk_balance()
     block = mk_block()
     callvalue = mk_callvalue()
-    caller = mk_caller(solver)
-    this = mk_this(solver)
+    caller = mk_addr('msg_sender')
+    this = mk_addr('this_address')
 
     sevm = SEVM(options)
     ex = sevm.mk_exec(
@@ -237,7 +223,7 @@ def setup(
 
     solver = mk_solver(args)
 
-    this = mk_this(solver)
+    this = mk_addr('this_address')
 
     sevm = SEVM(options)
 
@@ -249,7 +235,7 @@ def setup(
         block     = mk_block(),
         calldata  = [],
         callvalue = con(0),
-        caller    = mk_caller(solver),
+        caller    = mk_addr('msg_sender'),
         this      = this,
         symbolic  = False,
         solver    = solver,
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -159,7 +159,7 @@ def mk_block() -> Block:
 
 # TODO: addresses are used as keys in the context and could be treated as 160-bit values
 def mk_addr(name: str) -> Word:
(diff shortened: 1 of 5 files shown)


src/halmos/sevm.py
```diff
@@ -877,6 +921,26 @@ def call_unknown() -> None:
...
+                    # code must be concrete
+                    try:
+                        code_offset = extract_bytes(arg, 4 + 32, 32).as_long()
+                        code_length = extract_bytes(arg, 4 + code_offset, 32).as_long()
+                        code_int = extract_bytes(arg, 4 + code_offset + 32, code_length).as_long()
+                        code_bytes = code_int.to_bytes(code_length, 'big')
+
+                        pgm, code = decode_bytes(code_bytes)
+                        ex.pgm[who] = pgm
+                        ex.code[who] = code
```
daejunpark, line 938: in theory, this could be unsound when `who` is symbolic, because it may alias with other existing address keys.  note that pgm and code are plain python dictionaries, not smt arrays, so no alias reasoning is done here.

the same issue exists even if `who` is concrete, as the `this` address is currently symbolic and may alias with it. perhaps, we should make the `this` address concrete, so that we can avoid alias reasoning at least for concrete `who`.

tests/test/Foundry.t.sol
```diff
@@ -27,4 +27,28 @@ contract FoundryTest is Test {
...
         assertEq(counter2.cnt(), x);
     }
+
+    function testEtchConcrete() public {
+        vm.etch(address(0x42), hex"60425f526001601ff3");
+        (bool success, bytes memory retval) = address(0x42).call("");
+
+        assertTrue(success);
+        assertEq(retval.length, 1);
+        assertEq(uint256(uint8(retval[0])), 0x42);
```
daejunpark, line 37: just curious, what's the meaning of the return value here?  i mean, what does the code do?

tests/test/Foundry.t.sol
```diff
@@ -27,4 +27,28 @@ contract FoundryTest is Test {
...
+    }
+
+    function testEtchSymbolicAddr(address who) public {
+        vm.etch(who, hex"60425f526001601ff3");
+        (bool success, bytes memory retval) = who.call("");
+
+        assertTrue(success);
+        assertEq(retval.length, 1);
+        assertEq(uint256(uint8(retval[0])), 0x42);
```
daejunpark, line 46: suppose you create a new contract `c` at the beginning of this function, and add another assertion here saying that the code of `c` is unchanged, then it shouldn't pass because `who` may be equal to `c`, but the current version will not catch that, as i mentioned earlier.

src/halmos/__main__.py
```diff
@@ -189,8 +175,8 @@ def run_bytecode(hexcode: str, args: argparse.Namespace, options: Dict) -> List[
     balance = mk_balance()
     block = mk_block()
     callvalue = mk_callvalue()
-    caller = mk_caller(solver)
-    this = mk_this(solver)
+    caller = mk_addr('msg_sender')
+    this = mk_addr('this_address')
```
daejunpark, line 179: nice!

src/halmos/sevm.py
```diff
@@ -69,6 +71,29 @@ def wstore_bytes(mem: List[Byte], loc: int, size: int, arr: List[Byte]) -> None:
...
+    return val
+
+
+def extract_funsig(calldata: BitVecRef):
+    '''Extracts the function signature (first 4 bytes) from calldata'''
+    n = calldata.size()
+    # return simplify(Extract(n-1, n-32, calldata))
+    return extract_bytes(calldata, 0, 4)
+
+
```
daejunpark, line 96: great!!!

src/halmos/byte2op.py
```diff
@@ -35,7 +40,7 @@ def concat(args):
         return args[0]
 
 # Decode ByteCodes to Opcodes
-def decode(hexcode: Any) -> Tuple[List[Opcode], List[Any]]:
+def decode(hexcode: BitVecVal) -> Tuple[List[Opcode], List[Any]]:
```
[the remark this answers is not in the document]
karmacoma-eth: ah yes sorry, what's the most generic type we can use then? Would `BitVecRef` work?

tests/test/Foundry.t.sol
karmacoma-eth: it just returns 0x42, just wanted some "proof" that the code ran and returned the expected value

src/halmos/byte2op.py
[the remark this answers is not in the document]
daejunpark: yes, i think so: https://github.com/Z3Prover/z3/blob/f928b4460665410c83fbe2db2f202cba868fec56/src/api/python/z3/z3.py#L3936-L3947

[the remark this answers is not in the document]
daejunpark: btw, BitVecVal is a function that returns a BitVecNumRef object:
https://github.com/Z3Prover/z3/blob/f928b4460665410c83fbe2db2f202cba868fec56/src/api/python/z3/z3.py#L4012-L4026

but, `is_bv_value()` doesn't simply check if it is an instance of `BitVecNumRef`, so i don't think BitVecNumRef is a good type for bitvector values, either:
https://github.com/Z3Prover/z3/blob/f928b4460665410c83fbe2db2f202cba868fec56/src/api/python/z3/z3.py#L3950-L3962

tests/test/Foundry.t.sol
daejunpark: i see.  i was wondering whether it would mean the address `0x42` used in `vm.etch()`, or not.

src/halmos/sevm.py
```diff
@@ -48,7 +48,9 @@ def wextend(mem: List[Byte], loc: int, size: int) -> None:
 
 def wload(mem: List[Byte], loc: int, size: int) -> Bytes:
     wextend(mem, loc, size)
-    return simplify(Concat(mem[loc:loc+size])) # BitVecSort(size * 8)
+
+    # BitVecSort(size * 8)
+    return mem[loc] if size == 1 else simplify(Concat(mem[loc:loc+size]))
```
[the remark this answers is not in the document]
karmacoma-eth: in c535f6dce

src/halmos/sevm.py
karmacoma-eth: disabled in f76f1841d

src/halmos/byte2op.py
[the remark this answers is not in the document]
karmacoma-eth: BitVecRef annotation in 016355c20

daejunpark (approved):

PR #75, opened 2023-05-15
feat: add a new option that allows overwrite bytecode after setup

daejunpark:
A new option `--reset-bytecode 'addr1=code1,addr2=code2,...'` is added that resets the bytecode of the given addresses after setUp(). This can be useful when you want to run the same test with the actually deployed bytecode.

For example, the following test will pass if the bytecode of `c` is changed to the one where foo() returns 2, e.g.,: `halmos --reset-bytecode '8719=0x6080604052348015600f57600080fd5b506004361060285760003560e01c8063c298557814602d575b600080fd5b600260405190815260200160405180910390f3fea2646970667358221220c2880ecd3d663c2d8a036163ee7c5d65b9a7d1749e1132fd8ff89646c6621d5764736f6c63430008130033'`:

```
contract C {
    function foo() public returns (uint) {
        return 1;
    }
} 

contract ResetTest {
    C c;
    
    function setUp() public {
        c = new C();
    }

    function checkFoo() public {
        assert(c.foo() == 2);
    }

}
```

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -47,6 +47,7 @@ def parse_args(args=None) -> argparse.Namespace:
     parser.add_argument('--function', metavar='FUNCTION_NAME_PREFIX', default='check', help='run tests matching the given prefix only (default: %(default)s)')
 
     parser.add_argument('--bytecode', metavar='HEX_STRING', help='execute the given bytecode')
+    parser.add_argument('--reset-bytecode', metavar='ADDR1=CODE1,ADDR2=CODE2,...', help='reset the bytecode of given addresses after setUp()')
 
     parser.add_argument('--loop', metavar='MAX_BOUND', type=int, default=2, help='set loop unrolling bounds (default: %(default)s)')
     parser.add_argument('--width', metavar='MAX_WIDTH', type=int, help='set the max number of paths')
@@ -594,6 +595,14 @@ def main() -> int:
                         print(color_warn(f'Error: {setup_sig} failed: {type(err).__name__}: {err}'))
                         if args.debug: traceback.print_exc()
                         continue
+
+                    if args.reset_bytecode:
+                        for assign in [x.split('=') for x in args.reset_bytecode.split(',')]:
+                            addr = con(int(assign[0].strip(), 0))
+                            new_hexcode = assign[1].strip()
+                            (new_pgm, new_code) = decode_hex(new_hexcode)
+                            setup_ex.pgm[addr] = new_pgm
+                            setup_ex.code[addr] = new_code
 
                     for funsig in funsigs:
                         funselector = methodIdentifiers[funsig]
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -46,36 +46,53 @@ def parse_args(args=None) -> argparse.Namespace:
     parser.add_argument('--contract', metavar='CONTRACT_NAME', help='run tests in the given contract only')
     parser.add_argument('--function', metavar='FUNCTION_NAME_PREFIX', default='check', help='run tests matching the given prefix only (default: %(default)s)')
 
-    parser.add_argument('--bytecode', metavar='HEX_STRING', help='execute the given bytecode')
-    parser.add_argument('--reset-bytecode', metavar='ADDR1=CODE1,ADDR2=CODE2,...', help='reset the bytecode of given addresses after setUp()')
-
     parser.add_argument('--loop', metavar='MAX_BOUND', type=int, default=2, help='set loop unrolling bounds (default: %(default)s)')
     parser.add_argument('--width', metavar='MAX_WIDTH', type=int, help='set the max number of paths')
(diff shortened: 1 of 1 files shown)


src/halmos/__main__.py
```diff
@@ -29,6 +29,7 @@ def parse_args(args) -> argparse.Namespace:
     parser.add_argument('--function', metavar='FUNCTION_NAME_PREFIX', default='check', help='run tests matching the given prefix only (default: %(default)s)')
 
     parser.add_argument('--bytecode', metavar='HEX_STRING', help='execute the given bytecode')
+    parser.add_argument('--reset-bytecode', metavar='ADDR1=CODE1,ADDR2=CODE2,...', help='reset the bytecode of given addresses after setUp()')
```
karmacoma-eth, line 32: we wouldn't expect this to be used by "regular" end-users, right? 

maybe we can create an `advanced` or `internal` argument group:

```py
group_internal = parser.add_argument_group("Internal options")
group_internal.add_argument('--reset-bytecode', metavar='ADDR1=CODE1,ADDR2=CODE2,...', help='reset the bytecode of given addresses after setUp()')
```

src/halmos/__main__.py
```diff
@@ -566,6 +567,14 @@ def main() -> int:
                         if args.debug: traceback.print_exc()
                         continue
 
+                    if args.reset_bytecode:
+                        for assign in [x.split('=') for x in args.reset_bytecode.split(',')]:
+                            addr = con(int(assign[0].strip(), 0))
```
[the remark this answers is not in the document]
daejunpark: not sure if i correctly understood your comment, but you can pass an address in hex too (or in binary or octal), like `--reset-bytecode '0x220f=0x....'` because of [base=0](https://docs.python.org/3/library/functions.html#int).

src/halmos/__main__.py
daejunpark: sounds great, now i divided options to multiple groups. now it looks like:
```
$ halmos -h
usage: halmos [-h] [--contract CONTRACT_NAME] [--function FUNCTION_NAME_PREFIX] [--loop MAX_BOUND] [--width MAX_WIDTH] [--depth MAX_DEPTH] [--array-lengths NAME1=LENGTH1,NAME2=LENGTH2,...] [--symbolic-storage] [-v] [--debug] [--log LOG_FILE_PATH] [--print-revert]
              [--help-compile] [--no-smt-add] [--no-smt-sub] [--no-smt-mul] [--smt-div] [--smt-div-by-const] [--smt-mod-by-const] [--smt-exp-by-const N] [--solver-timeout-branching TIMEOUT] [--solver-timeout-assertion TIMEOUT] [--solver-fresh] [--solver-subprocess]
              [--bytecode HEX_STRING] [--reset-bytecode ADDR1=CODE1,ADDR2=CODE2,...] [--symbolic-jump] [--print-potential-counterexample]
              [TARGET_DIRECTORY]

positional arguments:
  TARGET_DIRECTORY      source root directory (default: current directory)

options:
  -h, --help            show this help message and exit
  --contract CONTRACT_NAME
                        run tests in the given contract only
  --function FUNCTION_NAME_PREFIX
                        run tests matching the given prefix only (default: check)
  --loop MAX_BOUND      set loop unrolling bounds (default: 2)
  --width MAX_WIDTH     set the max number of paths
  --depth MAX_DEPTH     set the max path length
  --array-lengths NAME1=LENGTH1,NAME2=LENGTH2,...
                        set the length of dynamic-sized arrays including bytes and string (default: loop unrolling bound)
  --symbolic-storage    set default storage values to symbolic

Debugging options:
  -v, --verbose         increase verbosity levels: -v, -vv, -vvv, -vvvv
  --debug               run in debug mode
  --log LOG_FILE_PATH   log individual execution steps in JSON
  --print-revert        print reverting paths in verbose mode

Build options:
  --help-compile        print build options (foundry, hardhat, etc.)

Solver options:
  --no-smt-add          do not interpret `+`
  --no-smt-sub          do not interpret `-`
  --no-smt-mul          do not interpret `*`
  --smt-div             interpret `/`
  --smt-div-by-const    interpret division by constant
  --smt-mod-by-const    interpret constant modulo
  --smt-exp-by-const N  interpret constant power up to N (default: 2)
  --solver-timeout-branching TIMEOUT
                        set timeout (in milliseconds) for solving branching conditions (default: 1000)
  --solver-timeout-assertion TIMEOUT
                        set timeout (in milliseconds) for solving assertion violation conditions (default: 60000)
  --solver-fresh        run an extra solver with a fresh state for unknown
  --solver-subprocess   run an extra solver in subprocess for unknown

Internal options:
  --bytecode HEX_STRING
                        execute the given bytecode
  --reset-bytecode ADDR1=CODE1,ADDR2=CODE2,...
                        reset the bytecode of given addresses after setUp()

Experimental options:
  --symbolic-jump       support symbolic jump destination
  --print-potential-counterexample
                        print potentially invalid counterexamples

For more information, see https://github.com/a16z/halmos
```

PR #77, opened 2023-05-16
Hide crytic-compile options by default

karmacoma-eth:
Use a 2-tier argument parsing system:
- first, extract the arguments known to halmos and use that to generate the usage/help message
- second, extract the arguments known to crytic-compile
- error out if there are arguments unknown to both

Recap:
- `halmos --help` prints only the halmos options (--loop, --function, --symbolic-storage, etc.)
- `halmos --compile-help` prints the crytic-compile options (--skip-clean, --ignore-compile, etc.)
- `halmos --nonsense` prints an error message

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -22,9 +22,22 @@ if hasattr(sys, 'set_int_max_str_digits'): # Python verion >=3.8.14, >=3.9.14, >
     sys.set_int_max_str_digits(0)
 
 def mk_crytic_parser() -> argparse.ArgumentParser:
-    crytic_compile_parser = argparse.ArgumentParser(prog='crytic-compile')
+    crytic_compile_parser = argparse.ArgumentParser()
     cryticparser.init(crytic_compile_parser)
     return crytic_compile_parser
+
+def print_help_compile(crytic_compile_parser: argparse.ArgumentParser) -> None:
+    formatter = crytic_compile_parser._get_formatter()
+    for action_group in crytic_compile_parser._action_groups:
+        if action_group.title == 'options':
+            # prints "--help", which is not helpful
+            continue
+
+        formatter.start_section(action_group.title)
+        formatter.add_text(action_group.description)
+        formatter.add_arguments(action_group._group_actions)
+        formatter.end_section()
+    crytic_compile_parser._print_message(formatter.format_help())
 
 def parse_args(args=None) -> argparse.Namespace:
     parser = argparse.ArgumentParser(prog='halmos', epilog='For more information, see https://github.com/a16z/halmos')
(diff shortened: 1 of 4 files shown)


src/halmos/__main__.py
```diff
@@ -21,7 +21,12 @@
 if hasattr(sys, 'set_int_max_str_digits'): # Python verion >=3.8.14, >=3.9.14, >=3.10.7, or >=3.11
     sys.set_int_max_str_digits(0)
 
-def parse_args(args) -> argparse.Namespace:
+def mk_crytic_parser() -> argparse.ArgumentParser:
+    crytic_compile_parser = argparse.ArgumentParser(prog='crytic-compile')
```
daejunpark, line 25: Which program name do we want to show in the --compile-help message: `crytic-compile` vs `halmos`?  The current PR version prints as follows:
```
$ halmos tests --compile-help
usage: crytic-compile [-h] [--compile-force-framework COMPILE_FORCE_FRAMEWORK] ...
       ^^^^^^^^^^^^^^
```

src/halmos/__main__.py
```diff
@@ -56,10 +61,9 @@ def parse_args(args) -> argparse.Namespace:
     parser.add_argument('--log', metavar='LOG_FILE_PATH', help='log individual execution steps in JSON')
     parser.add_argument('--print-revert', action='store_true', help='print reverting paths in verbose mode')
     parser.add_argument('--print-potential-counterexample', action='store_true', help='print potentially invalid counterexamples')
+    parser.add_argument('--compile-help', action='store_true', help='print build options (foundry, hardhat, etc.)')
```
daejunpark, line 64: `--compile-help` vs `--help-compile`?

daejunpark (commented):
This is great, thank you so much!

src/halmos/__main__.py
karmacoma-eth: You're right, that usage string is out of place. I removed it from the output in 5383aa6:

```
halmos --help-compile 2>&1 | head                                                                                                                                                                                                              

Compile options:
  --compile-force-framework COMPILE_FORCE_FRAMEWORK
                        Force the compile to a given framework (solc,truffle,e
                        mbark,dapp,etherlime,etherscan,vyper,waffle,brownie,so
                        lc-json,buidler,hardhat,foundry,standard,archive)
  --compile-remove-metadata
                        Remove the metadata from the bytecodes
  --compile-custom-build COMPILE_CUSTOM_BUILD
                        Replace platform specific build command
  --ignore-compile      Do not run compile of any platform
```

src/halmos/__main__.py
karmacoma-eth: `--help-compile` in 5383aa6 👍

daejunpark (approved):

PR #83, opened 2023-05-20
Refactor: Contract class to abstract over concrete/symbolic/mixed code

karmacoma-eth:
Highlights:

* `Contract` class can be instantiated from concrete `bytes` object or mixed Z3 expression 
* `Instruction` class replaces `Opcode`
* `Contract` supports random byte access, slicing, instruction decoding, iterating over instructions
* replaces decode, decode_hex, ops_to_pgm, Exec.code
* no more Exec.pgm

Why?

* does not automatically create a Z3 object for each concrete byte in the bytecode
* encapsulates the representation of the bytecode so that it's not spread all over the SEVM logic
* provides a single abstraction over `pgm` and `code`
* should be easier to serialize (either a single `bytes` object or a single `BitVecRef` expression)


karmacoma-eth:
Need to debug some issues, I have an example that fails with `ValueError: No successful path found in {setup_sig}` but it works on the main branch

daejunpark (approved):

daejunpark:
@karmacoma-eth I added some refactoring + fixes. Could you please take a look?

PR #88, opened 2023-05-26
perf: use concrete this and msg.sender

daejunpark:
concrete addresses:
- msg.sender: 0xaaaa0000
- this: 0xaaaa0001
- new addresses: 0xaaaa0002, 0xaaaa0003, ...

added optional flag `--symbolic-msg-sender` to have symbolic msg.sender.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -189,7 +189,6 @@ def mk_block() -> Block:
     block.chainid = con(1) # for ethereum
     return block
 
-# TODO: addresses are used as keys in the context and could be treated as 160-bit values
 def mk_addr(name: str) -> Word:
     return BitVec(name, 160)
 
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -51,6 +51,7 @@ def parse_args(args=None) -> argparse.Namespace:
     parser.add_argument('--array-lengths', metavar='NAME1=LENGTH1,NAME2=LENGTH2,...', help='set the length of dynamic-sized arrays including bytes and string (default: loop unrolling bound)')
 
     parser.add_argument('--symbolic-storage', action='store_true', help='set default storage values to symbolic')
+    parser.add_argument('--symbolic-msg-sender', action='store_true', help='set msg.sender symbolic')
 
     # debugging options
     group_debug = parser.add_argument_group("Debugging options")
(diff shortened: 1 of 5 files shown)


src/halmos/__main__.py
```diff
@@ -192,6 +193,15 @@ def mk_block() -> Block:
 def mk_addr(name: str) -> Word:
     return BitVec(name, 160)
 
+def mk_caller(args: argparse.Namespace) -> Word:
+    if args.symbolic_msg_sender:
+        return mk_addr('msg_sender')
+    else:
+        return con_addr(999)
+
+def mk_this() -> Word:
```
karmacoma-eth, line 202: they should return `Address` now, right?

src/halmos/__main__.py
```diff
@@ -192,6 +193,15 @@ def mk_block() -> Block:
 def mk_addr(name: str) -> Word:
     return BitVec(name, 160)
 
+def mk_caller(args: argparse.Namespace) -> Word:
+    if args.symbolic_msg_sender:
+        return mk_addr('msg_sender')
+    else:
+        return con_addr(999)
```
karmacoma-eth, line 200: not a big fan of 999 and 1000, for no really good reason other than they don't look like addresses 😅

I would prefer either of these:
- using well known values like the default msg.sender and test address used in foundry
- or if we don't want to couple tightly with foundry (their values may change, etc, etc), we could come up with new "well known values", e.g.:
```
➜ address(uint160(uint256(keccak256("halmos default msg.sender"))))
Type: address
└ Data: 0x6215e5fd0f1d85fb78fb8e27cede12699bfebb5a
➜ address(uint160(uint256(keccak256("halmos default address(this)"))))
Type: address
└ Data: 0x7d025cd0feaea4e87a75c716e3ed70279df723bd
```
(they're static but don't really look like anything in logs)
- or using "magic values" that pop in logs like the dolphin address (`0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`)

karmacoma-eth (dismissed):
minor suggestions otherwise makes sense to me

src/halmos/__main__.py
daejunpark: makes sense, but i'd like to have them to be smaller numbers since it makes it easier to assume they don't collide with other sha3 hash values generated at runtime. (we already have sha3 values are not very close to 0 or 2^256.)

would some hex values look like addresses?  like `0x1000` or `0xbeef0001`?

PR #89, opened 2023-05-26
perf: convert rawcode to bytes if concrete

daejunpark:
codecopy becomes slow if rawcode is bitvector. this is a temporary workaround.


karmacoma-eth (approved):
Makes sense! 👍

For the moment, we still store only bitvectors in Exec.mem, right? I can see 2 other ways to improve:
- if the whole range to load in memory is concrete bv values, build bytes directly rather than `Concat(bv_values...)` followed by `rawcode.as_long().to_bytes(rawcode.size() // 8, 'big')`
- maybe longer term, but I think it would be even nicer to have mem store a mix of ints and bitvectors, so deploying a concrete contract would not involve z3 at all

PR #93, opened 2023-05-30
feat: parallel processing of assertions

karmacoma-eth:
Needs to be enabled with `--solver-parallel` for now, might be enabled by default later on

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -9,10 +9,11 @@ import argparse
 import re
 import traceback
 
-from timeit import default_timer as timer
-
 from crytic_compile import cryticparser
 from crytic_compile import CryticCompile, InvalidCompilation
+from dataclasses import dataclass
+from multiprocessing import Pool
+from timeit import default_timer as timer
 
 from .utils import color_good, color_warn
 from .sevm import *
@@ -82,6 +83,8 @@ def parse_args(args=None) -> argparse.Namespace:
     group_solver.add_argument('--solver-timeout-assertion', metavar='TIMEOUT', type=int, default=1000, help='set timeout (in milliseconds) for solving assertion violation conditions (default: %(default)s)')
     group_solver.add_argument('--solver-fresh', action='store_true', help='run an extra solver with a fresh state for unknown')
     group_solver.add_argument('--solver-subprocess', action='store_true', help='run an extra solver in subprocess for unknown')
+    group_solver.add_argument('--solver-parallel', action='store_true', help='run assertion solvers in parallel')
+    group_solver.add_argument('--solver-parallel-cores', default=os.cpu_count(), type=int, help='max number of cores to use for parallel assertion solvers (default: %(default)s)')
 
     # internal options
     group_internal = parser.add_argument_group("Internal options")
@@ -235,12 +238,11 @@ def run_bytecode(hexcode: str, args: argparse.Namespace, options: Dict) -> List[
     )
     (exs, _) = sevm.run(ex)
 
-    models = []
     for idx, ex in enumerate(exs):
         opcode = ex.current_opcode()
         if opcode in [EVM.STOP, EVM.RETURN, EVM.REVERT, EVM.INVALID]:
-            gen_model(args, models, idx, ex)
-            print(f'Final opcode: {mnemonic(opcode)} | Return data: {ex.output} | Input example: {models[-1][0]}')
+            model_with_context = gen_model(args, idx, ex)
+            print(f'Final opcode: {mnemonic(opcode)} | Return data: {ex.output} | Input example: {model_with_context.model}')
         else:
             print(color_warn(f'Not supported: {mnemonic(opcode)} {ex.error}'))
         if args.verbose >= 1:
@@ -315,6 +317,14 @@ def setup(
         print(f'[time] setup: {setup_end - setup_start:0.2f}s (decode: {setup_mid - setup_start:0.2f}s, run: {setup_end - setup_mid:0.2f}s)')
 
     return setup_ex
+
+
+@dataclass(frozen=True)
+class ModelWithContext:
+    model: UnionType[Model, str]
+    index: int
+    result: CheckSatResult
+
 
 def run(
     setup_ex: Exec,
@@ -393,8 +403,10 @@ def run(
 
     # check assertion violations
     normal = 0
-    models = []
+    execs_to_model = []
+    models: List[ModelWithContext] = []
     stuck = []
+
     for idx, ex in enumerate(exs):
         if args.debug: print(f'Checking output: {idx+1} / {len(exs)}')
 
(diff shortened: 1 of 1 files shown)


karmacoma-eth:
without `--solver-parallel` on ERC721A:
```[time] total: 975.49s (build: 1.27s, tests: 974.22s)```

with `--solver-parallel`: 
```[time] total: 491.18s (build: 1.29s, tests: 489.89s)```

biggest improvement is this absolute unit of a test:
```[PASS] testTransferOtherOwnershipPreservation(address,address,uint256,uint256) (paths: 442/1354, time: 601.07s (paths: 225.07s, models: 376.00s), bounds: [])```

becomes
```
[PASS] testTransferOtherOwnershipPreservation(address,address,uint256,uint256) (paths: 438/1345, time: 248.13s (paths: 197.99s, models: 50.15s), bounds: [])
```

File: src/halmos/__main__.py [hunk 1]
```diff
@@ -393,38 +403,50 @@ def run(
...
+                execs_to_model.append((idx, ex))
         elif ex.failed:
-            gen_model(args, models, idx, ex)
+            if args.debug: print(f'  Will generate model for failed execution')
+            execs_to_model.append((idx, ex))
         else:
             stuck.append((opcode, idx, ex))
 
-    end = timer()
+    if len(execs_to_model) > 2 and args.solver_parallel:
```
daejunpark, line 428: why not parallel when len is 2?  would the overhead be bigger than saving?

src/halmos/__main__.py
```diff
@@ -393,38 +403,50 @@ def run(
...
-    passed = (normal > 0 and len(models) == 0 and len(stuck) == 0)
-    if passed:
-        passfail = color_good('[PASS]')
     else:
-        passfail = color_warn('[FAIL]')
+        models = [gen_model(args, idx, ex) for idx, ex in execs_to_model]
+
+    end = timer()
+
+    passed = (normal > 0 and any(m.model is not None for m in models) == 0 and len(stuck) == 0)
```
daejunpark, line 438: there may exist `m` such that m.model is None but m.result is unknown.  in that case, i don't think we want to say it is passed.

btw, doesn't `any()` return boolean?  i mean, `False == 0` is true, but not easy to read for me.

src/halmos/__main__.py
```diff
@@ -433,12 +455,12 @@ def run(
             elif args.print_potential_counterexample:
                 print(color_warn(f'Counterexample (potentially invalid): {str_model(model, args)}'))
             else:
-                print(color_warn(f'Counterexample: unknown'))
-        else:
-            print(color_warn(f'Counterexample: unknown'))
+                print(color_warn(f'Counterexample: {result}'))
```
daejunpark, line 458: this doesn't print anything when model is None and result is unknown.  is it intended?

src/halmos/__main__.py
karmacoma-eth: will rewrite, the expression is confusing

src/halmos/__main__.py
karmacoma-eth: no, we probably want to warn

src/halmos/__main__.py
karmacoma-eth: should read better in 3fe2666

src/halmos/__main__.py
```diff
@@ -393,38 +403,51 @@ def run(
...
-        passfail = color_good('[PASS]')
     else:
-        passfail = color_warn('[FAIL]')
+        models = [gen_model(args, idx, ex) for idx, ex in execs_to_model]
+
+    end = timer()
+
+    no_counterexample = all(m.model is None for m in models)
+    passed = (no_counterexample and normal > 0 and len(stuck) == 0)
+    passfail = color_good('[PASS]') if passed else color_warn('[FAIL]')
```
daejunpark, line 440: `no_counterexample` should check if all results are unsat, rather than all models are None.

as i mentioned earlier, e.g., if we have a single normal path, and a single panic(1) path for which counterexample generation failed due to timeout, then we shouldn't consider it as passed.

src/halmos/__main__.py
```diff
@@ -433,12 +456,14 @@ def run(
             elif args.print_potential_counterexample:
                 print(color_warn(f'Counterexample (potentially invalid): {str_model(model, args)}'))
             else:
-                print(color_warn(f'Counterexample: unknown'))
-        else:
-            print(color_warn(f'Counterexample: unknown'))
+                print(color_warn(f'Counterexample: {result}'))
```
daejunpark, line 459: at this point, `result` will be always `sat`, but we want to print a different message, because the generated counterexample is potentially invalid. this is because it includes an interpretation of `evm_div` (uninterpreted abstraction of bitvector division); see `is_valid_model()`.

so i suggest something like "counterexample is unknown", or "no valid counterexample", or more detailed explanation.

src/halmos/__main__.py
```diff
@@ -393,38 +403,50 @@ def run(
...
+            if args.debug: print(f'  Will generate model for failed execution')
+            execs_to_model.append((idx, ex))
         else:
             stuck.append((opcode, idx, ex))
 
-    end = timer()
+    if len(execs_to_model) > 2 and args.solver_parallel:
+        with Pool(processes=args.solver_parallel_cores) as pool:
+            if args.debug: print(f'Spawning {len(execs_to_model)} parallel assertion solvers on {args.solver_parallel_cores} cores')
+            models = [m for m in pool.starmap(gen_model_from_sexpr, [(args, idx, ex.solver.sexpr()) for idx, ex in execs_to_model])]
```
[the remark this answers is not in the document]
daejunpark: good idea. but i think we should still save the other smt queries as well.  this way, if the first counterexample is not interesting or practically infeasible, users can still check the remaining queries to find better counterexamples.

src/halmos/__main__.py
karmacoma-eth: I kinda disagree on this one, as a user of halmos I don't want to see a test `[FAIL]` without a valid counterexample I can work with. Maybe a warning that I should increase the assert solver timeout, but it's not a conclusive fail. Which is fine since we're doing bounded model checking, no?

src/halmos/__main__.py
karmacoma-eth: Yes, in fact I think we should do something like slither:

```
C.destination (test.sol#3) is never initialized. It is used in:
	- f (test.sol#5-7)
Reference: https://github.com/trailofbits/slither/wiki/Vulnerabilities-Description#uninitialized-state-variables
```

or errorprone:

```
$ bazel build :hello
ERROR: example/myproject/BUILD:29:1: Java compilation in rule '//example/myproject:hello'
ShortSet.java:6: error: [CollectionIncompatibleType] Argument 'i - 1' should not be passed to this method;
its type int is not compatible with its collection's type argument Short
      s.remove(i - 1);
              ^
    (see http://errorprone.info/bugpattern/CollectionIncompatibleType)
1 error
```
 
👉 display a short but readable error message, with a machine-readable error-code (`uninitialized-state-variables`, `CollectionIncompatibleType`) that can be used to link to a longer explanation

src/halmos/__main__.py
daejunpark: make sense. warning would be a better way to convey the result. in that case, i think it would be useful to provide the unsolved queries so that users can directly check them (potentially modifying the queries and/or using different solvers) without re-runing the entire tests.

for this PR, simple warning messages would suffice. we can work on providing queries later.

PR #97, opened 2023-06-02
perf: use to_smt2() instead of sexpr() for assertion solving

daejunpark:
not entirely sure about this, but may be worth considering.

in certain cases, sexpr() is significantly slower than to_smt2(), but haven't seen the opposite yet.


karmacoma-eth (approved):
if it works let's go for it 👍

speaking of which, looks like CI is not running for this PR? 🤨

daejunpark:
> speaking of which, looks like CI is not running for this PR? 🤨

that's because it's not based on main.

karmacoma-eth:
> that's because it's not based on main.

ah makes sense, do we want to keep it this way though?

daejunpark:
> > that's because it's not based on main.
> 
> ah makes sense, do we want to keep it this way though?

yeah let's enable tests for this as well.  perhaps, it's good time to review the overall settings and update things that can be improved?  would you be able to take a look with fresh mind?

PR #101, opened 2023-06-06
feat: support ADDMOD, MULMOD

daejunpark:
Encoding:
- `ADDMOD(x, y, z)`: `Extract(255, 0, (ZeroExt(  1, x) + ZeroExt(  1, y)) % ZeroExt(  1, z))`
- `MULMOD(x, y, z)`: `Extract(255, 0, (ZeroExt(256, x) * ZeroExt(256, y)) % ZeroExt(256, z))`

Note that the operands are converted to a larger bitsize (257-bit for addition, and 512-bit for multiplication) to avoid overflow in the intermediate operation, and finally casted down to 256-bit.

Also, `--smt-mod` option is added to support for full mod reasoning.

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -934,12 +934,12 @@ class SEVM:
         w2 = b2i(w2)
         w3 = b2i(w3)
         if op == EVM.ADDMOD:
-            r1 = self.arith(ex, EVM.ADD, ZeroExt(1, w1), ZeroExt(1, w2)) # to avoid add overflow
-            r2 = self.arith(ex, EVM.MOD, r1, ZeroExt(1, w3))
+            r1 = self.arith(ex, EVM.ADD, simplify(ZeroExt(1, w1)), simplify(ZeroExt(1, w2))) # to avoid add overflow
+            r2 = self.arith(ex, EVM.MOD, simplify(r1), simplify(ZeroExt(1, w3)))
             return Extract(255, 0, r2)
         elif op == EVM.MULMOD:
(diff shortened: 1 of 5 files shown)


src/halmos/sevm.py
```diff
@@ -1529,6 +1557,9 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps]:
                 elif EVM.ADD <= opcode <= EVM.SMOD: # ADD MUL SUB DIV SDIV MOD SMOD
                     ex.st.push(self.arith(ex, opcode, ex.st.pop(), ex.st.pop()))
 
+                elif EVM.ADDMOD <= opcode <= EVM.MULMOD: # ADDMOD MULMOD
```
karmacoma-eth, line 1560: we always interpret these regardless of `--smt-mod`, right?

daejunpark: symbolic mod operations will be uninterpreted, unless `--smt-mod` is provided.

concrete mod operations will be always interpreted, yes.

the following cases are considered as "concrete operations":
- when both operands are concrete
- mod by 0 or 1 (which can be reduced to 0)
- mod by a power of 2 (which can be reduced to lower-bit extraction)
- if `--smt-mod-by-const` is provided

--

the same thing also happens for division operations along with `--smt-div` and `--smt-div-by-const`.

daejunpark: see also #103, which replaces the uninterpreted abstraction with the actual symbols for assertion solving in external process.

this way, we can avoid unnecessary nonlinear reasoning for path exploration, and later do full nonlinear reasoning when generating counterexamples.

PR #102, opened 2023-06-06
feat: warning for incomplete path exploration due to loop bound

daejunpark:


src/halmos/__main__.py
```diff
@@ -293,7 +293,11 @@ def setup(
         dyn_param_size = [] # TODO: propagate to run
         mk_calldata(abi, setup_name, setup_sig, arrlen, args, setup_ex.calldata, dyn_param_size)
 
-        (setup_exs_all, setup_steps) = sevm.run(setup_ex)
+        (setup_exs_all, setup_steps, setup_bounded_loops) = sevm.run(setup_ex)
+
+        if setup_bounded_loops:
+            warn(LOOP_BOUND, f'{setup_sig}: paths have not been fully explored due to the loop unrolling bound: {args.loop}')
```
karmacoma-eth, line 299: nice use of the new error codes :)

daejunpark: yes, it's quite convenient, thank you!!

PR #105, opened 2023-06-08
Faster codecopy + create

karmacoma-eth:
Before this PR:
- everything is lowered to a bitvector byte before being stored in memory

After this PR:
- memory is now an array of either concrete or bitvector bytes
- doing a codecopy of concrete bytes keeps the bytes concrete in memory
-   (furthermore, doing a codecopy of a concrete *portion* of symbolic bytecode will also keep these bytes concrete)
- doing a create using concrete bytes in memory keeps the bytecode of the new contract concrete
- for fully symbolic bytecode or memory, the behavior is unchanged

The result is that large concrete bytecode creation is now orders of magnitude faster, which is a very common case in realistic test suites. This saves tens of thousands of Extract and Concat operation on every test!

*Caveat*: it's hard to guarantee that there aren't weird side effects (we lose the consistency of always symbolic bytes in memory), we're probably due for a separate memory refactor and mixed concrete/symbolic value wrapper.

*Perf*: 
- ERC721A before: 8 min
- ERC721A after: 3min41

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -306,7 +306,7 @@ def setup(
                 if res != unsat:
                     setup_exs.append(setup_ex)
 
-        if len(setup_exs) == 0: raise ValueError('No successful path found in {setup_sig}')
+        if len(setup_exs) == 0: raise ValueError(f'No successful path found in {setup_sig}')
         if len(setup_exs) > 1:
             print(color_warn(f'Warning: multiple paths were found in {setup_sig}; an arbitrary path has been selected for the following tests.'))
             if args.debug: print('\n'.join(map(str, setup_exs)))
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -299,12 +299,15 @@ def setup(
             if args.debug: print('\n'.join(setup_bounded_loops))
 
         setup_exs = []
+
         for idx, setup_ex in enumerate(setup_exs_all):
             if setup_ex.current_opcode() in [EVM.STOP, EVM.RETURN]:
                 setup_ex.solver.set(timeout=args.solver_timeout_assertion)
                 res = setup_ex.solver.check()
                 if res != unsat:
                     setup_exs.append(setup_ex)
+            elif args.debug:
+                print(color_warn(f'setup execution finished with {mnemonic(setup_ex.current_opcode())} (error={setup_ex.error})'))
 
         if len(setup_exs) == 0: raise ValueError(f'No successful path found in {setup_sig}')
         if len(setup_exs) > 1:
@@ -749,7 +752,7 @@ def main() -> int:
     if args.statistics:
         print(f'\n[time] total: {main_end - main_start:0.2f}s (build: {main_mid - main_start:0.2f}s, tests: {main_end - main_mid:0.2f}s)')
 
-    if (total_passed + total_failed) == 0:
+    if not funsigs:
         error_msg = f'Error: No tests with the prefix `{args.function}`'
         if args.contract is not None:
             error_msg += f' in {args.contract}'
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -83,6 +83,7 @@ class Instruction:
 class NotConcreteError(Exception):
     pass
 
+
 def unbox_int(x: Any) -> Any:
     '''Convert int-like objects to int'''
     if isinstance(x, bytes):
@@ -93,6 +94,7 @@ def unbox_int(x: Any) -> Any:
 
     return x
 
+
 def int_of(x: Any, err: str = 'expected concrete value but got') -> int:
     res = unbox_int(x)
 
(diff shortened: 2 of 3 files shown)


src/halmos/__main__.py
```diff
@@ -299,14 +299,17 @@ def setup(
...
         setup_exs = []
+
         for idx, setup_ex in enumerate(setup_exs_all):
             if setup_ex.current_opcode() in [EVM.STOP, EVM.RETURN]:
                 setup_ex.solver.set(timeout=args.solver_timeout_assertion)
                 res = setup_ex.solver.check()
                 if res != unsat:
                     setup_exs.append(setup_ex)
+            elif args.debug:
+                print(color_warn(f'setup execution finished with {mnemonic(setup_ex.current_opcode())} (error={setup_ex.error})'))
```
karmacoma-eth, line 310: setup_ex errors were hidden, even with --debug

src/halmos/__main__.py
```diff
@@ -299,14 +299,17 @@ def setup(
...
 
-        if len(setup_exs) == 0: raise ValueError('No successful path found in {setup_sig}')
+        if len(setup_exs) == 0: raise ValueError(f'No successful path found in {setup_sig}')
```
karmacoma-eth, line 312: typo resulted in expression not being evaluated

src/halmos/__main__.py
```diff
@@ -749,7 +752,7 @@ def main() -> int:
     if args.statistics:
         print(f'\n[time] total: {main_end - main_start:0.2f}s (build: {main_mid - main_start:0.2f}s, tests: {main_end - main_mid:0.2f}s)')
 
-    if (total_passed + total_failed) == 0:
+    if not funsigs:
```
karmacoma-eth, line 755: errors running `setUp()` used to result in the misleading `Error: No tests with the prefix...` message

src/halmos/sevm.py
```diff
@@ -151,14 +183,21 @@ def instruction_length(opcode: Any) -> int:
     return (opcode - EVM.PUSH0 + 1) if EVM.PUSH1 <= opcode <= EVM.PUSH32 else 1
 
 def wextend(mem: List[Byte], loc: int, size: int) -> None:
-    if len(mem) < loc + size:
-        mem.extend([BitVecVal(0, 8) for _ in range(loc + size - len(mem))])
+    mem.extend([0] * (loc + size - len(mem)))
```
karmacoma-eth, line 186: optimization: expand memory with 0s instead of con(0)

src/halmos/sevm.py
```diff
@@ -151,14 +183,21 @@ def instruction_length(opcode: Any) -> int:
...
 
-def wload(mem: List[Byte], loc: int, size: int) -> Bytes:
+def wload(mem: List[Byte], loc: int, size: int, prefer_concrete=False) -> UnionType[bytes,Bytes]:
```
karmacoma-eth, line 188: prefer_concrete should hopefully be a temporary crutch -- I feel like this should be the default going forward but is a little risky now

src/halmos/__main__.py
karmacoma-eth: actually this is not quite right, `funsigs` may not have been assigned

daejunpark: ah you already fixed that! :)

src/halmos/sevm.py
```diff
@@ -101,16 +103,46 @@ def int_of(x: Any, err: str = 'expected concrete value but got') -> int:
...
+
+
+def iter_bytes(x: Any):
+    '''Return an iterable over the bytes of x (concrete or symbolic)'''
+
+    if isinstance(x, bytes):
+        return x
+
+    if isinstance(x, int):
+        return x.to_bytes((x.bit_length() + 7) // 8, 'big')
```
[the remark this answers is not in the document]
karmacoma-eth: yes I agree! That's why I try to stick to bytes and not int, I like that they carry their expected size

src/halmos/sevm.py
daejunpark: yeah, i also considered that, but i'm concerned that it might introduce too many unnecessary conversions between bv to int, back and forth, for some cases.  i think the current usage, limited to the codecopy-like cases, strikes a good balance.  let's expand its use when we encounter obvious cases that can improve performance!

src/halmos/sevm.py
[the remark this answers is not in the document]
daejunpark: i'd suggest either removing support for the int case altogether, or requiring the bitsize as an additional input, which will ensure that the int case is used only for intended purposes.

src/halmos/sevm.py
```diff
@@ -151,14 +183,21 @@ def instruction_length(opcode: Any) -> int:
...
+    mem.extend([0] * (loc + size - len(mem)))
 
-def wload(mem: List[Byte], loc: int, size: int) -> Bytes:
+def wload(mem: List[Byte], loc: int, size: int, prefer_concrete=False) -> UnionType[bytes,Bytes]:
     wextend(mem, loc, size)
 
+    memslice = mem[loc:loc+size]
+
+    if prefer_concrete and all(is_concrete(i) for i in memslice):
+        return bytes([int_of(i) for i in memslice])
```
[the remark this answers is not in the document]
karmacoma-eth: yes that's right, mem is now a list of either int or bv

src/halmos/sevm.py
```diff
@@ -151,14 +183,21 @@ def instruction_length(opcode: Any) -> int:
...
+
+    # wrap concrete bytes in BitVecs
+    wrapped = [BitVecVal(i, 8) if not is_bv(i) else i for i in mem[loc:loc+size]]
```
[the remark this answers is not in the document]
karmacoma-eth: thanks, good catch

src/halmos/sevm.py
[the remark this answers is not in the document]
karmacoma-eth: I added the required length argument for ints

src/halmos/sevm.py
```diff
@@ -146,33 +180,61 @@ def byte_length(x: Any) -> int:
...
+
+
+def wload(mem: List[UnionType[int, BitVecRef]], loc: int, size: int, prefer_concrete=False) -> UnionType[bytes,Bytes]:
     wextend(mem, loc, size)
 
+    memslice = mem[loc:loc+size]
+
+    # runtime sanity check: mem should only contain ints or BitVecs (not bytes)
+    if not all(isinstance(i, int) or is_bv(i) for i in memslice):
+        raise ValueError(memslice)
```
daejunpark, line 200: would it be possible to incorporate this check into the existing iteration below instead of having a separate scan?  or would it be too complicated, and performance gain wouldn't justify the effort?

daejunpark (approved):

PR #106, opened 2023-06-08
perf: not use z3 const arrays

daejunpark:
Previously z3 const arrays (which are not part of smtlib2) were used to model empty storage mappings.  A problem of using non-smtlib2 symbols was that solvers with explicit logic (e.g., QF_AUFBV) sometimes generated invalid counterexamples, because any smtlib2 logic solver invalidates the const arrays.  On the other hand, not specifying a logic wasn't an option either, because the auto-config solver sometimes became very slow.

To address these issues, the current approach no longer relies on z3 const arrays. Instead, regular arrays are used, with emptiness axioms for each array lookup key that arises during execution.

--

This PR also includes several minor improvements in counterexample outputs.


src/halmos/__main__.py
```diff
@@ -583,10 +583,8 @@ def select(var):
         elif args.verbose >= 1:
             if name.startswith('storage') or name.startswith('msg_') or name.startswith('this_'): return True
         return False
-    if args.debug:
-        return str(model)
-    else:
-        return '[' + ', '.join(sorted(map(lambda decl: f'{decl} = {model[decl]}', filter(select, model)))) + ']'
+    select_model = filter(select, model) if not args.debug else model
+    return '[' + f',\n{" "*17}'.join(sorted(map(lambda decl: f'{decl} = {hexify(model[decl])}', select_model))) + ']'
```
karmacoma-eth, line 587: 17 spaces? that's oddly specific

src/halmos/utils.py
```diff
@@ -1,9 +1,21 @@
...
 
 from z3 import *
 
+def hexify(x):
+    if isinstance(x, str):
+        return re.sub(r'\b(\d+)\b', lambda match: hex(int(match.group(1))), x)
+    elif isinstance(x, int):
+        return hex(x)
+    elif is_bv_value(x):
+        return hex(x.as_long())
```
karmacoma-eth, line 15: I think it would be great to preserve the bitsize of the value in the hex output

```
# not this:
hex(BitVecVal(0x42, 160).as_long())
'0x42'

# this:
'0x' +  BitVecVal(0x42, 160).as_long().to_bytes(20, 'big').hex()
'0x0000000000000000000000000000000000000042'
```

karmacoma-eth: Using bv_value_to_bytes from the other PR, I would suggest the following:

```py
def hexify(x):
    if isinstance(x, str):
        return re.sub(r'\b(\d+)\b', lambda match: hex(int(match.group(1))), x)
    elif isinstance(x, int):
        return hex(x)
    elif isinstance(x, bytes):
        return '0x' + x.hex()
    elif is_bv_value(x):
        return hexify(bv_value_to_bytes(x))
    else:
        return hexify(str(x))
```

src/halmos/__main__.py
daejunpark: ah, yeah, i was trying to print counterexamples like:
```
Counterexample: [x = 0x1,
                 y = 0x2,
                 z = 0x3]
Counterexample: [x = 0x11,
                 y = 0x22,
                 z = 0x33]
```

multi-line is used to better handle the cases with many variables (which happen in real examples).

please feel free to suggest if you have a better idea to print these!

src/halmos/utils.py
daejunpark: good point! will do!

karmacoma-eth (dismissed):

src/halmos/__main__.py
daejunpark: now it prints:
```
Counterexample: [
    x = 0x1,
    y = 0x2,
    z = 0x3]
```

PR #107, opened 2023-06-10
Rework parallel pool

karmacoma-eth:
Different take on parallel processing now that `setup()` runs faster: instead of running `setup()` once and sharing a serialized `setup_ex` with the worker process for each test, actually run `setup() + run()` for each test.

This also uses a shared ProcessPoolExecutor, since we can't have embedded multiprocessing.Pool. As a result, we should be able to combine `--solver-parallel` and `--test-parallel`

~⚠️ not ready to be merged, still experimental! (in particular the main process seems stuck at the end when we combine `--solver-parallel` and `--test-parallel`)~

*UPDATE*: with [b9d3704](https://github.com/a16z/halmos/pull/107/commits/b9d3704635610805d3abe813bc7a278b68afe41d), we can now combine `--solver-parallel` and `--test-parallel`

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -588,7 +588,7 @@ def run_parallel(run_args: RunArgs) -> Tuple[int, int]:
     hexcode, abi, methodIdentifiers = run_args.hexcode, run_args.abi, run_args.methodIdentifiers
 
     setup_info = extract_setup(methodIdentifiers)
-    if args.verbose >= 1:
+    if setup_info.sig and args.verbose >= 1:
         print(f'Running {setup_info.sig}')
 
     fun_infos = [FunctionInfo(funsig.split('(')[0], funsig, methodIdentifiers[funsig]) for funsig in run_args.funsigs]
@@ -605,7 +605,7 @@ def run_parallel(run_args: RunArgs) -> Tuple[int, int]:
 
 def run_sequential(run_args: RunArgs) -> Tuple[int, int]:
     setup_info = extract_setup(run_args.methodIdentifiers)
-    if args.verbose >= 1:
+    if setup_info.sig and args.verbose >= 1:
         print(f'Running {setup_info.sig}')
 
     try:
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -147,7 +147,7 @@ def find_abi(abi: List, fun_info: FunctionInfo) -> Dict:
             return item
     raise ValueError(f'No {funsig} found in {abi}')
 
-def mk_calldata(abi: List, fun_info: FunctionInfo, cd: List, dyn_param_size: List[str]) -> None:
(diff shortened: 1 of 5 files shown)


src/halmos/__main__.py
```diff
@@ -638,34 +794,17 @@ def main() -> int:
     # command line arguments
     #
 
+    global args
```
karmacoma-eth, line 797: I don't love introducing globals, but in this case I think it's reasonable because args is written to only once but read by basically function, which means that it pollutes every function parameter list

src/halmos/__main__.py
```diff
@@ -91,13 +97,13 @@ def parse_args(args=None) -> argparse.Namespace:
     group_solver.add_argument('--solver-fresh', action='store_true', help='run an extra solver with a fresh state for unknown')
     group_solver.add_argument('--solver-subprocess', action='store_true', help='run an extra solver in subprocess for unknown')
     group_solver.add_argument('--solver-parallel', action='store_true', help='run assertion solvers in parallel')
-    group_solver.add_argument('--solver-parallel-cores', default=os.cpu_count(), type=int, help='max number of cores to use for parallel assertion solvers (default: %(default)s)')
```
karmacoma-eth, line 94: I ended up removing it for a couple reasons:

- not sure if there's a good use case for limiting the number (I certainly would like 100% usage if I can)
- the interplay between processes and threads makes this somewhat meaningless? at least much harder to guarantee 
- pools can be instantiated in a different module without requiring arg parsing

src/halmos/__main__.py
```diff
@@ -504,7 +523,123 @@ def run(
...
+        print(f'Running {setup_info.sig}')
+
+    fun_infos = [FunctionInfo(funsig.split('(')[0], funsig, methodIdentifiers[funsig]) for funsig in run_args.funsigs]
+    single_run_args = [SetupAndRunSingleArgs(hexcode, abi, setup_info, fun_info, args) for fun_info in fun_infos]
+
+    # dispatch to the shared process pool
+    exitcodes = process_pool.map(setup_and_run_single, single_run_args)
+
+    num_passed = sum(1 for x in exitcodes if x == 0)
+    num_failed = sum(1 for x in exitcodes if x != 0)
```
daejunpark, line 601: at this point, `num_failed` is always 0, because `existcodes` becomes empty after the lazy iterables have already been consumed when computing `num_passed`.

you can reproduce this by running two tests in parallel where one of them fails.

karmacoma-eth: oh damn, great find 😱

fixed in 910a1cb

daejunpark (approved):

src/halmos/__main__.py
daejunpark: ok, let's improve this later if needed.

src/halmos/__main__.py
daejunpark: ok, let's come back to this later if needed.

Issue #115, opened 2023-07-06
Print version number with `--version`


msooseth:
**Is your feature request related to a problem? Please describe.**
I'd like to get the version of `halmos` so so when benchmarking using https://github.com/eth-sc-comp/benchmarks I can get the version programmatically. This would allow me to compare the performance of different `halmos` versions with ease

**Describe the solution you'd like**
A new option `--version` that prints the current version number. A git hash would also be nice, but not necessary.

**Describe alternatives you've considered**
I could get the version from nix I think. But it may be helpful to users to know their version number, too.

**Additional context**
I am trying to check the performance of different symbolic/concolic EVM execution engines.

daejunpark:
Thank you for the suggestion! It was on our todo list, but we will prioritize it more now!

daejunpark:
fixed by #117

PR #117, opened 2023-07-14
feat: add --version option

daejunpark:
simply print the installed halmos package version number, without separately maintaining the version info in the source.


daejunpark:
fix #115

PR #118, opened 2023-07-14
feat: new forge frontend

daejunpark:
reimplement forge frontend without crytic-compile dependency

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,6 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "crytic-compile >= 0.3.0, < 0.3.2",
     "z3-solver",
 ]
 dynamic = ["version"]
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -766,7 +766,7 @@ def parse_build_out(args: argparse.Namespace) -> Dict:
     result = {} # compiler version -> source filename -> contract name -> (json, type)
 
     out_path = os.path.join(args.root, args.forge_build_out)
-    if not os.path.exists(out_path): raise FileNotFoundError(f'{out_path} does not exist')
+    if not os.path.exists(out_path): raise FileNotFoundError(f'the build output directory `{out_path}` does not exist')
 
     for sol_dirname in os.listdir(out_path): # for each source filename
         if not sol_dirname.endswith('.sol'): continue
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -60,6 +60,11 @@ def parse_args(args=None) -> argparse.Namespace:
     group_debug.add_argument('--print-setup-states', action='store_true', help='print setup execution states')
     group_debug.add_argument('--print-full-model', action='store_true', help='print full counterexample model')
 
+    # build options
+    group_build = parser.add_argument_group("Build options")
+
+    group_build.add_argument('--forge-build-out', metavar='DIRECTORY_NAME', default='out', help='forge build artifacts directory name (default: %(default)s)')
+
     # smt solver options
     group_solver = parser.add_argument_group("Solver options")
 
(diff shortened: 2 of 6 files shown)


src/halmos/__main__.py
```diff
@@ -819,19 +836,25 @@ def main() -> int:
     # compile
     #
 
-    try:
-        crytic_compile_args, crytic_compile_unknown_args  = crytic_compile_parser.parse_known_args()
+    build_cmd = [
+        'forge', # shutil.which('forge')
```
daejunpark, line 840: not sure if we need `shutil.which('forge')` here, or not, especially in the windows env. what do you think, karma?

pyproject.toml
```diff
@@ -18,7 +18,6 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "crytic-compile >= 0.3.0, < 0.3.2",
```
karmacoma-eth, line 21: feels good man 💖

src/halmos/__main__.py
karmacoma-eth: I don't think so, this should be fine

src/halmos/__main__.py
```diff
@@ -819,19 +836,25 @@ def main() -> int:
...
+        '--root', args.root,
+        '--extra-output', 'storageLayout', 'metadata'
+    ]
 
-        both_unknown = set(halmos_unknown_args) & set(crytic_compile_unknown_args)
-        if both_unknown:
-            print(color_warn(f'error: unrecognized arguments: {" ".join(both_unknown)}'))
-            return 1
+    # run forge without capturing stdout/stderr
+    build_exitcode = subprocess.run(build_cmd).returncode
```
karmacoma-eth, line 847: I think it's nice that we see the compiler output now 👌

```
[⠔] Compiling...
[⠒] Compiling 1 files with 0.5.17
[⠘] Compiling 43 files with 0.8.19
[⠢] Solc 0.5.17 finished in 9.82ms
[⠒] Solc 0.8.19 finished in 3.40s
```

karmacoma-eth: (and it makes it easier to troubleshoot if there's a compilation error too)

src/halmos/__main__.py
```diff
@@ -819,19 +836,25 @@ def main() -> int:
...
-    except InvalidCompilation as e:
-        print(color_warn(f'Parse error: {e}'))
+    if build_exitcode:
+        print(color_warn(f'build failed: {build_cmd}'))
         return 1
 
+    try:
+        build_out = parse_build_out(args)
+    except Error as err:
+        print(color_warn(f'build output parsing failed: {err}'))
```
karmacoma-eth, line 856: I think there's a problem with the except block

```
    except Error as err:
NameError: name 'Error' is not defined. Did you mean: 'OSError'?
```

karmacoma-eth:
```suggestion
    except Exception as ex:
        print(color_warn(f'build output parsing failed: {ex}'))
```

karmacoma-eth: should we continue after this exception or abort? 

if we continue, it fails with 

```
    for compiler_version in sorted(build_out):
UnboundLocalError: local variable 'build_out' referenced before assignment
```

src/halmos/__main__.py
```diff
@@ -783,6 +762,49 @@ def mk_arrlen(args: argparse.Namespace) -> Dict[str, int]:
     return arrlen
 
 
+def parse_build_out(args: argparse.Namespace) -> Dict:
+    result = {} # compiler version -> source filename -> contract name -> (json, type)
+
+    out_path = os.path.join(args.root, args.forge_build_out)
+    if not os.path.exists(out_path): raise ValueError('not exist', out_path)
```
karmacoma-eth, line 769:
```suggestion
    if not os.path.exists(out_path): raise ValueError(f'{out_path} does not exist')
```

karmacoma-eth: alternatively, we may want to throw the built-in `FileNotFoundError`?

karmacoma-eth (dismissed):
minor comments otherwise LGTM! 👌

karmacoma-eth (dismissed):

src/halmos/__main__.py
daejunpark: yes, that was the intention.

src/halmos/warnings.py
```diff
@@ -5,6 +5,7 @@
 from .utils import color_warn
 
 logger = logging.getLogger(__name__)
+logging.basicConfig()
```
daejunpark, line 8: karma, this is needed for the basic formatting of logging messages with the prefix of "WARNING:halmos.warnings:".
so far, it was implicitly executed when importing crytic_compile. please let me know if you don't want to keep this.

src/halmos/__main__.py
daejunpark: good catch, thanks!!

src/halmos/warnings.py
karmacoma-eth: ah I see! no it's great, let's keep it

karmacoma-eth (approved):

PR #119, opened 2023-07-14
add black formatting

karmacoma-eth:
bells and whistles:

- updated dev instructions in README.md
- pre-commit hook
- lint GitHub action 🧑‍⚖️
- mega refactor commit added to [.git-blame-ignore-revs](https://github.com/a16z/halmos/pull/119/commits/aecf3d8d040d33fad4bd978537df02e262b8b788)


tests/test_sevm.py
```diff
@@ -68,7 +99,9 @@ def test_run(hexcode, stack, pc, opcode: int, sevm, solver, storage):
     assert ex.pc == pc
     assert ex.current_opcode() == int_of(opcode)
 
+
 def byte_of(i, x):
+    # fmt: off
```
karmacoma-eth, line 104: this is the only customization I made

daejunpark: let's not make an exception, as it is subjective to determine when it is acceptable.

there are other places where the formatted code is very ugly.  we can work on them in separate prs by rewriting them in a formatting-friendly way.

README.md
```diff
@@ -79,6 +79,37 @@ Counterexample: [p_price_uint96 = 39614081294025656978550816768, p_quantity_uint
...
+# if you want to submit a pull request, fork the repository:
+
+gh repo fork a16z/halmos
+
+# or if you just want to develop locally, clone it
+# git clone git@github.com:a16z/halmos.git
+
+# create and activate a virtual environment
+
+python3.10 -m venv .venv
```
daejunpark, line 94: 3.10 doesn't seem representative.  we support >=3.8 and the latest is 3.11.  can we just say `python3`?

karmacoma-eth: let's recommend 3.11 then, makes more sense to me

tests/test_sevm.py
daejunpark: ah, i will take back. this is a test file, and i couldn't think of a good way to write it in the format friendly way (it needs to be verbose to cross-check with the main code logic.)

karmacoma-eth: image

daejunpark (approved):
looks great, thanks!

tests/test_sevm.py
karmacoma-eth: yeah I was going to say I agree about avoiding exceptions, but given that this is a test file and the extreme nesting makes black spiral into madness, so I think an exception is fine 

image

daejunpark (approved):

PR #121, opened 2023-07-17
feat: switch to pytest for regression testing

daejunpark:


src/halmos/__main__.py
```diff
@@ -101,6 +101,9 @@ def parse_args(args=None) -> argparse.Namespace:
     group_debug.add_argument(
         "--log", metavar="LOG_FILE_PATH", help="log every execution steps in JSON"
     )
+    group_debug.add_argument(
+        "--json", metavar="JSON_FILE_PATH", help="output test results in JSON"
```
karmacoma-eth, line 105: nit: the name of the flag isn't very descriptive, maybe `--json-output` or something like that?

src/halmos/__main__.py
```diff
@@ -532,12 +535,22 @@ class ModelWithContext:
     result: CheckSatResult
 
 
+@dataclass(frozen=True)
+class TestResult:
```
karmacoma-eth, line 539: nice!

src/halmos/__main__.py
```diff
@@ -790,7 +810,7 @@ class RunArgs:
     methodIdentifiers: Dict[str, str]
 
 
-def run_parallel(run_args: RunArgs) -> Tuple[int, int]:
+def run_parallel(run_args: RunArgs) -> Tuple[int, int, List[TestResult]]:
```
karmacoma-eth, line 813: it would feel cleaner to have the following signatures instead:

```
def run_parallel(run_args: RunArgs) -> List[TestResult]:
def run_sequential(run_args: RunArgs) -> List[TestResult]:
```

and the consumer can easily derive `num_passed` and `num_failed` from the `List[TestResult]`

tests/test/Counter.t.sol
```diff
@@ -1,30 +1,32 @@
 // SPDX-License-Identifier: AGPL-3.0
 pragma solidity >=0.8.0 <0.9.0;
 
+// NOTE: required options: --loop 4 --symbolic-storage
```
karmacoma-eth, line 4: out of scope for this but it would be really great if we could annotate it at the contract or method level

```
/// @custom:halmos loop=4
/// @custom:halmos symbolic-storage=true
```

src/halmos/__main__.py
```diff
@@ -1195,8 +1224,19 @@ def main() -> int:
                     total_passed += num_passed
                     total_failed += num_failed
 
+                    if test_results_map is not None:
+                        if contract_path in test_results_map:
+                            raise ValueError("already exists", contract_path)
+                        test_results_map[contract_path] = list(
```
karmacoma-eth, line 1230: arbitrary nit, I prefer the comprehension version:

`test_results_map[contract_path] = [asdict(t) for t in test_results]`

src/halmos/__main__.py
```diff
@@ -1089,7 +1114,7 @@ def parse_build_out(args: argparse.Namespace) -> Dict:
     return result
 
 
-def main() -> int:
+def main(argv=None, test_results_map=None) -> int:
```
karmacoma-eth, line 1117: it seems a little odd to pass a "write argument" to main, but ok 🙃 could use a docstring explaining that it's basically for testing, to essentially avoid going through file IO right?

daejunpark: i wasn't happy with this either, but did so to avoid unnecessary `asdict` conversion for test results in normal execution.  i changed the code to return the test results.

src/halmos/__main__.py
daejunpark: good idea! done.

tests/test/Counter.t.sol
daejunpark: yes, it would be great!  it's on the todo list.

karmacoma-eth (approved):
Looking good! Excited to get more robust testing in place 💪

PR #124, opened 2023-07-19
chore: tweak counterexample formatting

karmacoma-eth:
Before:

    Counterexample: [
        p_x_uint256 = 0x69,
        p_y_uint256 = 0x69]

After:

    Counterexample:
        p_x_uint256 = 0x69
        p_y_uint256 = 0x69

The unbalanced brackets are creating a tension that stays with me all day


daejunpark:
i like the version without the brackets, and actually that's what i initially tried, but i remember that an issue is when the filtered list is empty.  in that case, it looks like the following, and i remember it was confusing with the unknown result (maybe no longer the case).
```
Counterexample:
```

karmacoma-eth:
> when the filtered list is empty

mhhh I see, it does happen for no-input symbolic tests, and with my change the `model` string is empty, there it fails the `if model` test and it displays this now:

```
[FAIL] check_invariant() (paths: 0/2, time: 0.43s, bounds: [])
WARNING:Halmos:Counterexample: sat
```

Maybe in this case it should just display this line:

```
[FAIL] check_invariant() (paths: 0/2, time: 0.43s, bounds: [])
```

wdyt?

src/halmos/__main__.py
```diff
@@ -1002,17 +1002,11 @@ def is_valid_model(model) -> bool:
...
-        return False
+        return name.startswith("p_") or name.startswith("halmos_")
 
     select_model = filter(select, model) if not args.print_full_model else model
-    result = ",".join(
-        sorted(map(lambda decl: f"\n    {decl} = {hexify(model[decl])}", select_model))
-    )
-    return f"[{result}]"
+    formatted = [f"\n    {decl} = {hexify(model[decl])}" for decl in select_model]
+    return "".join(sorted(formatted))
```
daejunpark, line 1009: can we return some meaningful value (or message) when the return value is empty string, so that it can be printed out like this?
```
Counterexample: <some-value-or-message>
```

i believe it is useful info, telling users that no specific inputs are involved with this assertion violation.

daejunpark: for example, how about returning `"[]"` (without the leading newline) when the filtered list is empty?

karmacoma-eth: how about something cute like this? :)

```
[FAIL] check_invariant() (paths: 0/2, time: 0.43s, bounds: [])
Counterexample: ∅
```

daejunpark: amazing!

daejunpark: ah, there are some issues with handling the unicode empty set symbol `\u2205` in windows:
```
______________________________ test_run_bytecode ______________________________

args = Namespace(root='D:\\a\\halmos\\halmos', contract=None, function='check', loop=2, width=None, depth=None, array_lengths...lse, bytecode=None, reset_bytecode=None, test_parallel=False, symbolic_jump=True, print_potential_counterexample=False)

    def test_run_bytecode(args):
        # sets the flag in the global args for the main module
        args.symbolic_jump = True
    
        hexcode = "34381856FDFDFDFDFDFD5B00"
>       exs = run_bytecode(hexcode)

tests\test_cli.py:117: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src\halmos\__main__.py:424: in run_bytecode
    print(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <encodings.cp1252.IncrementalEncoder object at 0x000001A402AC6750>
input = 'Final opcode: STOP | Return data: None | Input example: \u2205'
final = False

    def encode(self, input, final=False):
>       return codecs.charmap_encode(input,self.errors,encoding_table)[0]
E       UnicodeEncodeError: 'charmap' codec can't encode character '\u2205' in position 56: character maps to <undefined>

C:\hostedtoolcache\windows\Python\3.11.4\x64\Lib\encodings\cp1252.py:19: UnicodeEncodeError
```

karmacoma-eth: got 'em with [edf439f](https://github.com/a16z/halmos/pull/124/commits/edf439f640d70961d0bf6139252d91b9cc9f4619)

image

daejunpark (approved):

PR #133, opened 2023-07-20
feat: support natspec annotations

### LINKED ISSUE #131: Support for custom NatSpec annotations
## Problem

Suppose you have a codebase with multiple test contracts:
- you can run the tests in ContractA with the default settings
- one test in ContractB requires `--loop 10`
- one test in ContractB requires `--smt-div`

In order for halmos to run the entire test suite, you resort to running `halmos --loop 10 --smt-div` which makes the tests in `ContractA` 10x slower.

## Solution

It would be amazing if we could support something like [custom NatSpec tags](https://docs.soliditylang.org/en/v0.8.21/natspec-format.html#tags) at the contract and function level, hypothetical example:

```solidity
/// @custom:halmos --loop 10
contract B {
    function check_loops() public {
        ...
    }

    /// @custom:halmos --smt-div
    function check_loopsAndDiv() public {
        ...
    }
}
```

(not sure about the exact syntax we want -- just a suggestion)

The main point is that this would enable these settings only for the relevant functions and contracts, not the entire test suite.

So in our hypothetical example, you could just run `halmos`, and:
- the tests in ContractA would run fast with the default settings
- all the tests in ContractB would run with `--loop 10`
- only `check_loopsAndDiv` would run with `--loop 10 --smt-div`

daejunpark:
fix #131 

Support `@custom:halmos` natspec annotation to specify additional commandline options.

The natspec for a contract is inherited by every function in the contract. The natspec for a function can overwrite the inherited natspec.

For example, given:
```
/// @custom:halmos --opt1 --loop 1
contract C {

    /// @custom:halmos --opt2 --loop 2
    function checkA() { ... }
}
```
The options for the `checkA()` function are: `--opt1 --opt2 --loop 2`

Note that it inherits `--opt1` but overwrites `--loop 1`.

--

The halmos natspec can be provided in multiline:
```
/// @custom:halmos --x
///                --y
```
or, using multiple tags:
```
/// @custom:halmos --x
/// @custom:halmos --y
```


tests/expected/all.json
```diff
@@ -653,8 +775,8 @@
         "test/Reset.t.sol:ResetTest": [
             {
                 "name": "checkFoo()",
-                "exitcode": 1,
```
karmacoma-eth, line 656: ah so the test used to fail, but now with the custom halmos annotation it passes? Amazing 🥳

daejunpark: yes, exactly.

src/halmos/__main__.py
```diff
@@ -887,9 +700,18 @@ def run_sequential(run_args: RunArgs) -> List[TestResult]:
     return test_results
 
 
+def extend_args(args: Namespace, more_opts: str) -> Namespace:
+    if more_opts:
+        new_args = deepcopy(args)
```
karmacoma-eth, line 705: I was looking for something to avoid the need for a deepcopy here, I ended up with this:

```python
import argparse
import collections

parser = argparse.ArgumentParser()
parser.add_argument('--foo')
parser.add_argument('--bar')
parser.add_argument('--loop')

def dictify(args):
    return {k: v for k, v in vars(args).items() if v is not None}

defaults = parser.parse_args(['--foo', 'beep', '--bar', 'boop'])
overrides = parser.parse_args(['--foo', 'ding', '--loop', '42'])
combined = collections.ChainMap(dictify(overrides), dictify(defaults))

print(combined['foo']) # 'ding' comes from overrides, takes precedence over defaults
print(combined['bar']) # 'boop' comes from defaults
print(combined['loop']) # '42' comes from overrides, not specified in defaults
```

ChainMap is cool but in the end it's not necessarily better because there is still some copying involved because of a poor interaction with Namespace. Might make sense to write our own `NamespaceChain`-like wrapper that avoids any copying at all at a later point

src/halmos/__main__.py
```diff
@@ -1124,14 +947,36 @@ def parse_build_out(args: argparse.Namespace) -> Dict:
     return result
 
 
+def parse_devdoc(funsig: str, contract_json: Dict) -> str:
+    try:
+        return contract_json["metadata"]["output"]["devdoc"]["methods"][funsig][
+            "custom:halmos"
+        ]
+    except Exception as err:
```
karmacoma-eth, line 955: catching Exception here as a shorthand because any of these keys might be missing, right?

karmacoma-eth: might want to catch KeyError to avoid overcatching

daejunpark: yes

src/halmos/__main__.py
daejunpark: do you suggest using chainmap now, or doing something better later?

i wasn't super happy with using deepcopy here, but thought that the namespace object should be small enough and deepcopy wouldn't affect performance much. are there any other issues that i missed?

karmacoma-eth: no it's just a general dislike of deepcopy creep. Disregard for now

karmacoma-eth (approved):
this is great, can't wait to start using it

daejunpark:
the external test failures are independent from this pr. it will be fixed by #134. will merge this pr after merging that.

PR #135, opened 2023-07-21
feat: add basic runner for baolean/symexec-bench

karmacoma-eth:
basic runner to make it easier to compare perf numbers across halmos versions

### DIFF
--- a/benchmarks/baolean.sh
+++ b/benchmarks/baolean.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env sh
+
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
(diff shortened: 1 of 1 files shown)


karmacoma-eth:
Could use a couple more bells and whistles:
- it clones in the current working directory, which makes it dirty wrt git. Might be better to checkout the external repo in a well defined location that is added to .gitignore?
- it would be nice to check that we get the expected output for a given test, otherwise the perf number is meaningless but not sure where to specify that

daejunpark (approved):
good start. we can further organize benchmarking stuffs when we discuss the eth-sc-comp.

PR #143, opened 2023-07-24
fix: revert state

### LINKED ISSUE #109: Not reverting storage states
**Describe the bug**
Failed to revert storage states.

_[Reported by Piotr at Halborn]_

**To Reproduce**
```
contract C {
  uint256 public num;
  function set1(uint256 x) public {
    num = x;
    revert("blah");
  }
}

contract CTest {
    function checkRevert1(uint256 x) public {
        C c = new C();
        require(x != 0);
        (bool result, ) = address(c).call(abi.encodeWithSignature("set1(uint256)", x));
        assert(!result);
        assert(c.num() != x); // fail
    }
}
```

daejunpark:
fix #109

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -1261,6 +1261,11 @@ class SEVM:
 
         caller = ex.prank.lookup(ex.this, to)
 
+        orig_code = ex.code.copy()
+        orig_storage = deepcopy(ex.storage)
+        orig_balance = deepcopy(ex.balance)
+        orig_log = deepcopy(ex.log)
+
         if not (is_bv_value(fund) and fund.as_long() == 0):
             ex.balance_update(
                 caller, self.arith(ex, EVM.SUB, ex.balance_of(caller), fund)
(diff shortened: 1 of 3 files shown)


src/halmos/sevm.py
```diff
@@ -1261,6 +1261,11 @@ def call(
 
         caller = ex.prank.lookup(ex.this, to)
 
+        orig_code = ex.code.copy()
+        orig_storage = deepcopy(ex.storage)
+        orig_balance = deepcopy(ex.balance)
+        orig_log = deepcopy(ex.log)
```
karmacoma-eth, line 1267: note to self: I have an idea to get rid of all these deepcopies, I'll get on it soon

tests/test/Revert.t.sol
```diff
@@ -37,4 +43,21 @@ contract CTest {
         assert(!result);
         assert(c.num() != x);
     }
+
+    function check_RevertBalance(bool paused, uint256 amount) public {
```
karmacoma-eth, line 47: cool! do we also have a check to verify that `code` is properly reverted too?

karmacoma-eth: ah I see https://github.com/a16z/halmos/issues/145 👀

karmacoma-eth (approved):

tests/test/Revert.t.sol
daejunpark: #145 is a different one.  i added a test for reverting code info.

karmacoma-eth (approved):
👌

PR #153, opened 2023-07-31
add getting started guide

daejunpark:


docs/getting-started.md
```diff
@@ -0,0 +1,260 @@
...
+```solidity
+// conceptual effect of symbolic testing of `check_transfer()`
+for (uint160 sender = 0; sender < type(uint160).max; sender++) {
+    for (uint160 receiver = 0; receiver < type(uint160).max; receiver++) {
+        for (uint256 amount = 0; amount < type(uint256).max; amount++) {
+            check_transfer(address(sender), address(receiver), amount);
+        }
+    }
+}
+```
```
karmacoma-eth, line 146: this is neat!

docs/getting-started.md
```diff
@@ -0,0 +1,260 @@
...
+vm.assume(receiver != address(0));
+vm.assume(token.balance(sender) >= amount);
+```
+Like fuzz tests, any input combinations that don't satisfy the `assume()` conditions are disregarded. This means that, after executing the above `assume()` statements, only the input combinations in which the receiver is non-zero and the sender has sufficient balance are considered. Other input combinations that violate these conditions are ignored.
+
+**Tips:**
+
+- You need to be careful not to exclude valid inputs by setting too strong input conditions.
+
+- In symbolic tests, avoid using `bound()` as it tends to perform poorly. Instead, use `vm.assume()` which is more efficient and easier to read and write.
```
karmacoma-eth, line 190: I would give an example like this:

```
uint256 tokenId = svm.createSymbolicUint256("tokenId");

// ❌ don't do this
tokenId = bound(tokenId, 1, MAX_TOKEN_ID);

// ✅ do this
vm.assume(1 <= tokenId && tokenId <= MAX_TOKEN_ID);
```

unfortunaly the bad pattern is more concise than our recommended pattern, maybe we want to introduce `svm.bound(val, min, max)`?

docs/getting-started.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+**Tips:**
+
+- If your goal is to check whether the target contract reverts under the expected conditions, a low-level call should be used. This allows the execution to continue even if the external call fails. Below is an example of a low-level call to the token transfer function. Note that the return value `success` can be subsequently used to check the reverting conditions.
+  ```solidity
+  vm.prank(sender);
+  (bool success,) = address(token).call(
+      abi.encodeWithSelector(token.transfer.selector, receiver, amount)
+  );
```
karmacoma-eth, line 209: do you want to add `assert(!success)` in the example?

that's something we could probably be doing by implementing `vm.expectRevert()`, right?

karmacoma-eth (approved):

PR #157, opened 2023-08-03
fix: test contract creation

### LINKED ISSUE #40: test contracts initialization
**Describe the bug**
Halmos doesn't properly initialize test contracts.

**To Reproduce**
```
contract FooTest {
  uint public value = 1;
  function testValue() public {
    assert(value == 1); // fail
  }
}
```

**Additional context**
Executing the contract creation code for test contracts is needed.

daejunpark:
fix #40

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -231,23 +231,18 @@ def run_bytecode(hexcode: str, args: Namespace) -> List[Exec]:
     return exs
 
 
-def setup(
+def deploy_test(
     hexcode: str,
-    abi: List,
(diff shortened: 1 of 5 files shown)


src/halmos/__main__.py
```diff
@@ -1073,7 +1105,7 @@ def _main(_args=None) -> MainResult:
                 if contract_type != "contract":
                     continue
 
-                hexcode = contract_json["deployedBytecode"]["object"]
```
karmacoma-eth, line 1076: small but significant change 😅

PR #159, opened 2023-08-04
feat: implement DELEGATECALL and CALLCODE

daejunpark:


tests/test/TestConstructor.t.sol
```diff
@@ -6,16 +6,26 @@ contract TestConstructorTest {
...
+    uint codesize_;
+    uint extcodesize_;
 
     constructor () {
         flag = 3;
         value = 4;
+
+        assembly {
+            sstore(codesize_.slot, codesize())
+            sstore(extcodesize_.slot, extcodesize(address()))
```
karmacoma-eth, line 18: ah interesting 👌

tests/test/Call.t.sol
```diff
@@ -0,0 +1,146 @@
...
+
+        assert(address(this).balance == 0);
+        assert(address(d).balance == fund);
+        assert(address(d.c()).balance == 0);
+    }
+
+    function check_delegatecall(uint x, uint fund) public payable {
+        vm.deal(address(this), fund);
+        vm.deal(address(d), 0);
+        vm.deal(address(d.c()), 0);
```
karmacoma-eth, line 107: why the deal 0s?

karmacoma-eth (approved):
looks great!

tests/test/Call.t.sol
daejunpark: balance is not assumed to be zero, because someone might have already sent fund to accounts before created.

PR #161, opened 2023-08-05
feat: support struct type

daejunpark:
fix #55

support struct and dynamic array types

now unsupported abi types are `fixedNxM`, `ufixedNxM`, and `function`

PR #162, opened 2023-08-05
feat: add --no-test-constructor option

daejunpark:
add a new option --no-test-constructor which doesn't run the constructor of test contracts and use the compiler-generated deployed bytecode.

this option is for backward compatibility.

this pr will be rebased after #161 is merged.


karmacoma-eth:
> this option is for backward compatibility

not sure I follow, do you have an example that needs this?

PR #163, opened 2023-08-06
feat: support CREATE2 

daejunpark:
add basic support for create2 opcode

note that halmos doesn't follow the keccak hash semantics when generating addresses for create/create2, (because symbolic hash computation is not practically feasible), but simply use a counter semantics, specifically:
- the addresses generated by create: 0xAAAA0001, 0xAAAA0002, 0xAAAA0003, ...
- the addresses generated by create2: 0xBBBB0000, 0xBBBB0001, 0xBBBB0002, ...

limitation: if salt or constructor arguments are symbolic, then the generated address depends on the name of symbols.  this means that if a create2 address is explicitly generated using different symbols with the same underlying value, then the resulting address will be considered different.

wip: do not merge

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -408,7 +408,7 @@ def run(
             #
             log=deepcopy(setup_ex.log),
             cnts=deepcopy(setup_ex.cnts),
-            sha3s=deepcopy(setup_ex.sha3s),
+            sha3s=deepcopy(setup_ex.sha3s),  # TODO: shallow copy
             storages=deepcopy(setup_ex.storages),
             balances=deepcopy(setup_ex.balances),
             calls=deepcopy(setup_ex.calls),
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -180,14 +180,14 @@ def run_bytecode(hexcode: str, args: Namespace) -> List[Exec]:
 
 
 def deploy_test(
-    hexcode: str,
+    creation_hexcode: str,
+    deployed_hexcode: str,
     sevm: SEVM,
     args: Namespace,
 ) -> Exec:
     # test contract creation bytecode
-    creation_bytecode = Contract.from_hexcode(hexcode)
+    creation_bytecode = Contract.from_hexcode(creation_hexcode)
 
-    solver = mk_solver(args)
     this = mk_this()
 
     ex = sevm.mk_exec(
@@ -201,8 +201,15 @@ def deploy_test(
         this=this,
         pgm=creation_bytecode,
         symbolic=False,
-        solver=solver,
+        solver=mk_solver(args),
     )
+
+    # use the given deployed bytecode if --no-test-constructor is enabled
+    if args.no_test_constructor:
+        deployed_bytecode = Contract.from_hexcode(deployed_hexcode)
+        ex.code[this] = deployed_bytecode
+        ex.pgm = deployed_bytecode
+        return ex
 
     # create test contract
     (exs, _, _) = sevm.run(ex)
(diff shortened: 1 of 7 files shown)


src/halmos/__main__.py
```diff
@@ -408,7 +408,7 @@ def run(
             #
             log=deepcopy(setup_ex.log),
             cnts=deepcopy(setup_ex.cnts),
-            sha3s=deepcopy(setup_ex.sha3s),
+            sha3s=deepcopy(setup_ex.sha3s),  # TODO: shallow copy
```
karmacoma-eth, line 411: wouldn't it make more sense to actually not copy sha3s at all and just hang on to the reference?

daejunpark: you mean reusing the sha3s dict for different tests?  no it doesn't work, because it will skip the no hash collision assumption for the existing hash expressions in the dict.

src/halmos/sevm.py
```diff
@@ -1747,12 +1757,23 @@ def create(
...
         # contract creation code
         create_hexcode = wload(ex.st.memory, loc, size, prefer_concrete=True)
         create_code = Contract(create_hexcode)
 
         # new account address
-        new_addr = ex.new_address()
+        if op == EVM.CREATE:
+            new_addr = ex.new_address()
+        else:  # EVM.CREATE2
+            code_hash = ex.sha3_data(create_hexcode, create_hexcode.size() // 8)
```
karmacoma-eth, line 1774: this fails when `create_hexcode` is concrete:

```
    code_hash = ex.sha3_data(create_hexcode, create_hexcode.size() // 8)
AttributeError: 'bytes' object has no attribute 'size'
```

```suggestion
            code_hash = ex.sha3_data(create_hexcode, byte_length(create_hexcode))
```

src/halmos/sevm.py
```diff
@@ -919,43 +921,50 @@ def normalize(expr: Any) -> Any:
...
 
-    def sha3_data(self, data: Bytes, size: int) -> None:
+    def sha3_data(self, data: Bytes, size: int) -> Word:
         f_sha3 = Function(
             "sha3_" + str(size * 8), BitVecSort(size * 8), BitVecSort(256)
         )
         sha3_expr = f_sha3(data)
-        sha3_output = BitVec(f"sha3_output_{len(self.sha3s):>02}", 256)
-        self.solver.add(sha3_output == sha3_expr)
+
```
karmacoma-eth, line 931: this also fails when `data` is concrete (`bytes`). Not very elegant but this wraps it up in a BV if needed:

```python
    def sha3_data(self, data: Bytes, size: int) -> Word:
        if not is_bv(data):
            if isinstance(data, bytes):
                data = int.from_bytes(data, "big")
            data = BitVecVal(data, size * 8)
```

tests/test/Create2.t.sol
```diff
@@ -0,0 +1,57 @@
...
+        set(x, y);
+    }
+
+    function set(uint x, uint y) public {
+        num1 = x;
+        num2 = y;
+    }
+}
+
+contract Create2Test is Test {
```
karmacoma-eth, line 20: cool, I would also add the following test to verify the collision behavior:

```
    function test_create2_collision() public {
        C c1 = new C{salt: bytes32(uint256(42))}(1, 2);
        C c2 = new C{salt: bytes32(uint256(42))}(1, 2);
    }
```

Ideally it would do a proper EVM revert, currently we leak an internal exception: (`ValueError: existing address: 3149594625`)

src/halmos/sevm.py
daejunpark: i'd like to minimize the mix uses of bv and bytes in the codebase.  it's not a long term solution anyway, and we need to implement a proper abstraction of symbolic values soon.

tests/test/Create2.t.sol
daejunpark: great suggestion for new test scenarios!  i added them.

src/halmos/sevm.py
karmacoma-eth: yes agreed

PR #165, opened 2023-08-06
feat: library linking

lowprivuser:
adds auto lib linking. ignores custom linking from CLI and `foundry.toml`


daejunpark (approved):
@lowprivuser thank you so much for implementing this much needed feature!  it looks great!  i've made some cleanups and adjusted the code to align with the latest master.

PR #166, opened 2023-08-10
Smol optimizations

karmacoma-eth:
2 things:
- cache the BitVecSort(n) invocations, which we do *a lot* of
- work with the natural stack/list ordering

Saves about 10% on some path-intensive benchmark (172M -> 156M function calls), coming mostly from caching

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -424,20 +424,16 @@ class State:
     def push(self, v: Word) -> None:
         if not (eq(v.sort(), BitVecSort256) or is_bool(v)):
             raise ValueError(v)
-        self.stack.insert(0, simplify(v))
+        self.stack.append(simplify(v))
 
     def pop(self) -> Word:
-        v = self.stack[0]
-        del self.stack[0]
-        return v
+        return self.stack.pop()
 
     def dup(self, n: int) -> None:
-        self.push(self.stack[n - 1])
+        self.push(self.stack[-n])
 
     def swap(self, n: int) -> None:
-        tmp = self.stack[0]
-        self.stack[0] = self.stack[n]
-        self.stack[n] = tmp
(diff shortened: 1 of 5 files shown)


src/halmos/sevm.py
```diff
@@ -31,43 +31,65 @@
...
+BitVecSorts = BitVecSortCache()
+
+# known, fixed BitVecSort sizes
+BitVecSort1 = BitVecSorts[1]
+BitVecSort8 = BitVecSorts[8]
+BitVecSort160 = BitVecSorts[160]
+BitVecSort256 = BitVecSorts[256]
+BitVecSort264 = BitVecSorts[264]
+BitVecSort512 = BitVecSorts[512]
+
```
daejunpark, line 55: Q: could you help me understand the behavior of this global mutable object when multiple SEVM objects are created?

for example, say we have two sevm objects:
```
from halmos.sevm import SEVM

sevm1 = SEVM(options)
sevm2 = SEVM(options)
```

will they have the same cache object, or a different one?

if the former, would it be safe when we run sevm1 and sevm2 in a thread or process pool?

in any case, can we avoid the use of global mutable objects?

karmacoma-eth: you're right I didn't think of that, both instances would access the same object because it's at the module level. I can make it read only to avoid this issue

daejunpark (approved):

karmacoma-eth:
I think the CI failures are now due to farcaster being in a bad state (https://github.com/farcasterxyz/contracts/pull/275#issuecomment-1673848974), unrelated to the changes in this PR

daejunpark:
yes, i know that.  cicada test was canceled because of that.  i set fail-fast false, so that we can make sure that all but farcaster test succeed.  let's merge after we confirm that.

PR #168, opened 2023-08-11
add NamedTimer class

karmacoma-eth:
+ reorganize main loop to reduce indentation level

The named timer makes it super easy to add a new subdivision (e.g. `timer.create_subtimer("parse")`) that will be automatically added to the report without having to do manual timing snapshots

### DIFF
--- a/.github/workflows/test-external.yml
+++ b/.github/workflows/test-external.yml
@@ -40,7 +40,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: halmos
-          submodules: recursive
+          # we won't be needing tests/lib for this workflow
+          submodules: false
 
       - name: Checkout external repo
         uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Install Foundry dependencies
         if: steps.foundry-cache.outputs.cache-hit != 'true'
-        run: forge install --shallow
+        run: forge install --root tests/ --shallow
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           submodules: false
 
-      - uses: actions/cache
+      - uses: actions/cache@v3
         id: foundry-cache
         with:
           path: tests/lib
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,20 +23,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          submodules: false
-
-      - uses: actions/cache@v3
-        id: foundry-cache
-        with:
-          path: tests/lib
-          key: cache-foundry-${{ hashFiles('.gitmodules') }}
+          submodules: recursive
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
-
-      - name: Install Foundry dependencies
-        if: steps.foundry-cache.outputs.cache-hit != 'true'
-        run: forge install --root tests/ --shallow
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,10 +23,20 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          submodules: recursive
+          submodules: false
+
+      - uses: actions/cache
+        id: foundry-cache
+        with:
+          path: tests/lib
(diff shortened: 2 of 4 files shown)


src/halmos/__main__.py
```diff
@@ -467,22 +464,20 @@ def run(
...
     passed = no_counterexample and normal > 0 and len(stuck) == 0
     if args.error_unknown:
         passed = passed and all(m.result == unsat for m in models)
     passfail = color_good("[PASS]") if passed else color_warn("[FAIL]")
 
-    time_total, time_paths, time_models = end - start, mid - start, end - mid
-    time_info = f"{time_total:0.2f}s"
-    if args.statistics:
-        time_info += f" (paths: {time_paths:0.2f}s, models: {time_models:0.2f}s)"
+    timer.stop()
```
karmacoma-eth, line 475: oops looks like I got a spurious stop() here

.github/workflows/test.yml
```diff
@@ -23,11 +23,21 @@ jobs:
...
       - uses: actions/checkout@v3
         with:
-          submodules: recursive
+          submodules: false
+
+      - uses: actions/cache@v3
+        id: foundry-cache
+        with:
+          path: tests/lib
+          key: cache-foundry-${{ hashFiles('.gitmodules') }}
```
karmacoma-eth, line 32: looks like this may be causing issues in the checkout post-job cleanup, but the tests still succeed 🤷‍♂️ 

```
/usr/bin/git submodule foreach --recursive sh -c "git config --local --name-only --get-regexp 'core\.sshCommand' && git config --local --unset-all 'core.sshCommand' || :"
fatal: not a git repository: tests/lib/forge-std/../../../.git/modules/tests/lib/forge-std
Warning: The process '/usr/bin/git' failed with exit code 128
```

daejunpark: my understanding is that the cache key needs to be unique for the specific snapshot of submodules.

the .gitmodules file itself doesn't include the refs hash for the submodules.  so even if existing submodules are updated to a newer version, the cache key will be the same, which will load the old submodules from the cache.

is there anything i'm missing here?

karmacoma-eth: yes this key could lead to a stale cache if the intent is to fetch the latest version of each dependency, since we don't have a lockfile equivalent AFAIK. This is just to experiment with caching and guaranteeing hits for now -- we could use a shorter lived key including for instance the date

daejunpark: i think the ci cache is nontrivial and deserves its own pr.  can we move this ci update to a separate pr?

src/halmos/__main__.py
daejunpark: when you stop() here, the end time of the previous subtimers ("paths" in this case) are reset again, which results in a wrong elapsed time for them.

src/halmos/utils.py
```diff
@@ -626,3 +627,68 @@ def mk_sha3_inv_offset(m: Dict[int, int]) -> Dict[int, Tuple[int, int]]:
...
+    def start(self):
+        if self.start_time is not None:
+            raise ValueError(f"Timer {self.name} has already been started.")
+        self.start_time = timer()
+
+    def stop(self, stop_subtimers=True):
+        if stop_subtimers:
+            for sub_timer in self.sub_timers:
+                sub_timer.stop()
+        self.end_time = timer()
```
daejunpark, line 648: i think the end_time should be set here only when it is None.

src/halmos/__main__.py
```diff
@@ -1025,6 +1018,14 @@ def import_libs(build_out_map: Dict, hexcode: str, linkReferences: Dict) -> Dict
...
 
 
+def build_output_iterator(build_out: Dict):
+    for compiler_version in sorted(build_out):
+        build_out_map = build_out[compiler_version]
+        for filename in sorted(build_out_map):
+            for contract_name in sorted(build_out_map[filename]):
+                yield (build_out_map, filename, contract_name)
+
+
```
daejunpark, line 1028: i like the refactoring made possible by introducing this.

daejunpark (changes requested):
please fix the subtimer issue, and move the ci update to a separate pr.

.github/workflows/test-external.yml
```diff
@@ -40,7 +40,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           path: halmos
-          submodules: recursive
+          # we won't be needing tests/lib for this workflow
+          submodules: false
```
karmacoma-eth, line 44: @daejunpark let's keep this change though right, I think it's good and inoffensive for external tests

karmacoma-eth:
> please fix the subtimer issue, and move the ci update to a separate pr.

in [0a8ae20](https://github.com/a16z/halmos/pull/168/commits/0a8ae202d2fa26317f19855090151233c8436ddf) and a211f88

daejunpark (approved):

PR #170, opened 2023-08-12
fix: improve unknown call resolution

daejunpark:
improve unknown external call resolution
- use semantic equality instead of structural equality to resolve known external call targets
- separately handle cheatcode and precompile calls
- do not allow unknown external calls, unless they are simple ether transfer (i.e., zero calldata)
- add options for assuming unknown calls to the given function signatures (e.g., isValidSignature(), onERC721Received(), etc) to be static
- fix ether transfer semantics bugs


daejunpark:
@karmacoma-eth now it is ready to review.

src/halmos/sevm.py
```diff
@@ -1323,6 +1330,27 @@ def arith2(self, ex: Exec, op: int, w1: Word, w2: Word, w3: Word) -> Word:
...
+        if target in ex.code:
+            return target
+
+        if target not in ex.alias:
+            for addr in ex.code:
+                if (
+                    # must equal
+                    ex.check(target != addr) == unsat
+                    # ensure existing path condition is feasible
+                    and ex.check(target == addr) != unsat
```
karmacoma-eth, line 1343: we take `unknown` as `may be feasible` here?

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
...
-        orig_balance = ex.balance
-        orig_log = ex.log.copy()
+        def send_callvalue(condition=None) -> None:
+            if (
+                # no balance update for CALLCODE which transfers to itself
+                op == EVM.CALL
+                and not (is_bv_value(fund) and fund.as_long() == 0)
+            ):
+                # assume balance is enough; otherwise ignore this path
+                ex.solver.add(UGE(ex.balance_of(caller), fund))
```
karmacoma-eth, line 1397: would it make sense to add this constraint to the `path` list or not really?

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
...
+
         caller = ex.prank.lookup(ex.this, to)
 
-        orig_code = ex.code.copy()
-        orig_storage = deepcopy(ex.storage)
-        orig_balance = ex.balance
-        orig_log = ex.log.copy()
+        def send_callvalue(condition=None) -> None:
+            if (
+                # no balance update for CALLCODE which transfers to itself
```
karmacoma-eth, line 1392: does that comment describe the code in the if? it's not obvious to me

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
...
-            # no balance update for CALLCODE which transfers to itself
-            ex.balance_update(
-                caller, self.arith(ex, EVM.SUB, ex.balance_of(caller), fund)
-            )
-            ex.balance_update(to, self.arith(ex, EVM.ADD, ex.balance_of(to), fund))
+                # conditional send
+                if condition is not None:
+                    value = If(condition, fund, con(0))
+                else:
+                    value = fund
```
karmacoma-eth, line 1403:
```suggestion
                value = If(condition, fund, con(0)) if condition is not None else fund
```

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
         if not ret_size >= 0:
             raise ValueError(ret_size)
 
+        if arg_size > 0:
+            arg = wload(ex.st.memory, arg_loc, arg_size)
+        else:
+            arg = None
+
```
karmacoma-eth, line 1387:
```suggestion
        arg = wload(ex.st.memory, arg_loc, arg_size) if arg_size > 0 else None
```

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
...
+                # conditional send
+                if condition is not None:
+                    value = If(condition, fund, con(0))
+                else:
+                    value = fund
+
+                ex.balance_update(
+                    caller, self.arith(ex, EVM.SUB, ex.balance_of(caller), value)
+                )
+                ex.balance_update(to, self.arith(ex, EVM.ADD, ex.balance_of(to), value))
```
karmacoma-eth, line 1408: it seems a little odd to dispatch to the ADD/SUB opcode handlers, is it to respect the --smt-add/--smt-sub options?

this would feel cleaner IMO:

```suggestion
                ex.balance_update(caller, ex.balance_of(caller) - value)
                ex.balance_update(to, ex.balance_of(to) + value)
```

src/halmos/sevm.py
```diff
@@ -1352,21 +1380,44 @@ def call(
...
+        def call_known(to: Address) -> None:
+            # backup current state
+            orig_code = ex.code.copy()
+            orig_storage = deepcopy(ex.storage)
+            orig_balance = ex.balance
+            orig_log = ex.log.copy()
 
-        def call_known() -> None:
+            # transfer msg.value
+            send_callvalue()
```
karmacoma-eth, line 1418: just parsing as I read through this: we send the value unconditionally before the call here here, does that mean that we later undo it if the call reverts?

src/halmos/sevm.py
```diff
@@ -1670,8 +1732,9 @@ def call_unknown() -> None:
                     store_account = uint160(Extract(767, 512, arg))
                     store_slot = simplify(Extract(511, 256, arg))
                     store_value = simplify(Extract(255, 0, arg))
-                    if store_account in ex.storage:
-                        ex.sstore(store_account, store_slot, store_value)
+                    store_account_addr = self.resolve_address_alias(ex, store_account)
```
karmacoma-eth, line 1735: 👍

src/halmos/sevm.py
```diff
@@ -1845,9 +1935,10 @@ def create(
         ex.storage[new_addr] = {}  # existing storage may not be empty and reset here
 
         # transfer value
-        # assume balance is enough; otherwise ignore this path
-        ex.solver.add(UGE(ex.balance_of(caller), value))
         if not (is_bv_value(value) and value.as_long() == 0):
+            # assume balance is enough; otherwise ignore this path
+            ex.solver.add(UGE(ex.balance_of(caller), value))
+
```
karmacoma-eth, line 1941: looks like this significantly duplicates the logic in `send_callvalue()`, maybe we want to factor this out independently? something like `transfer_value(from: Address, to: Address, value: BitVecRef) -> None`

src/halmos/sevm.py
daejunpark: yes.  not only here, but throughout the path exploration phase, we prune only unsat paths.

src/halmos/sevm.py
daejunpark: `path` is only for debugging purposes, so it doesn't really matter, but we can add this.

src/halmos/sevm.py
daejunpark: it explains why the msg.value transfer is done only for op == CALL, while CALLCODE also involves msg.value.

src/halmos/sevm.py
```diff
@@ -2247,20 +2340,30 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps]:
...
+                        codesize = con(len(ex.code[account_addr]))
                     else:
-                        codesize = f_extcodesize(address)
+                        codesize = f_extcodesize(account)
                         if (
-                            address == hevm_cheat_code.address
-                            or address == halmos_cheat_code.address
+                            eq(account, hevm_cheat_code.address)
+                            or eq(account, halmos_cheat_code.address)
+                            or eq(account, console.address)
```
karmacoma-eth, line 2354: do we need similar handling for address(0) here? Would it be helpful to add a constraint that it has no code? What about precompiles (I think addr <= 18)

src/halmos/sevm.py
daejunpark: yes, it is to handle the options.

src/halmos/sevm.py
daejunpark: yes

src/halmos/sevm.py
```diff
@@ -2247,20 +2340,30 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps]:
...
                     ex.st.push(codesize)
+                # TODO: define f_extcodehash for known addresses in advance
                 elif opcode == EVM.EXTCODEHASH:
-                    ex.st.push(f_extcodehash(ex.st.pop()))
+                    account = uint160(ex.st.pop())
+                    account_addr = self.resolve_address_alias(ex, account)
+                    if account_addr is not None:
+                        codehash = f_extcodehash(account_addr)
+                    else:
+                        codehash = f_extcodehash(account)
```
karmacoma-eth, line 2365: you know I'm a big fan of these expressions 🥰 

```suggestion
                    codehash = (
                        f_extcodehash(account_addr)
                        if account_addr is not None
                        else f_extcodehash(account)
                    )
```

karmacoma-eth: this is great btw, we're getting a ton of mileage out of the address aliases!

src/halmos/sevm.py
daejunpark: see todo comment. need more work around this. 

btw, why do we need special handling address(0), compared to other unknown addresses?

src/halmos/sevm.py
daejunpark: um, tbh, the ite expr is not readable to me if it doesn't fit in a single line.

src/halmos/sevm.py
daejunpark: added

src/halmos/sevm.py
daejunpark: refactored

src/halmos/sevm.py

src/halmos/sevm.py

src/halmos/sevm.py
daejunpark: will be done in a separate pr

karmacoma-eth: I'm not sure, intuitively I'm thinking that it may be useful to have special handling for address(0) as it is a "well known" address that for sure has no code

src/halmos/sevm.py
karmacoma-eth: readability isn't great but it's still cleaner IMO to have:
- a single assignment to codehash vs two
- the assignment be in scope of the use later

Issue #169, opened 2023-08-11
Counterexample in json output


zachwax101:
Would be great to have the counterexample for the failed test case in the extended json output

PR #171, opened 2023-08-12
fix: include model in json output

karmacoma-eth:
Problem: we would render and print models in the CLI, but they would not be included in the --json-output (even with --extended-json-output), which makes the json output not as useful as it could be.

Solution: the way we pass around models in the codebase is complicated, so I tried to keep a more structured dict around all the way into TestResults (instead of the formatted string). It can then either be included directly in the json or rendered on the CLI.

Fixes https://github.com/a16z/halmos/issues/169

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -20,6 +20,9 @@ from .warnings import *
 from .parser import mk_arg_parser
 from .calldata import Calldata
 
+StrModel = Dict[str, str]
+AnyModel = UnionType[Model, StrModel]
+
 arg_parser = mk_arg_parser()
 
 # Python version >=3.8.14, >=3.9.14, >=3.10.7, or >=3.11
@@ -334,8 +337,9 @@ def setup(
 
 @dataclass(frozen=True)
 class ModelWithContext:
-    model: str
-    validity: bool
+    # can be a filename containing the model or a dict with variable assignments
+    model: Optional[UnionType[StrModel, str]]
+    is_valid: Optional[bool]
     index: int
     result: CheckSatResult
 
@@ -345,6 +349,7 @@ class TestResult:
     name: str  # test function name
     exitcode: int  # 0: passed, 1: failed, 2: setup failed, ...
     num_models: int = None
+    models: List[ModelWithContext] = None
     num_paths: Tuple[int, int, int] = None  # number of paths: [total, success, blocked]
     time: Tuple[int, int, int] = None  # time: [total, paths, models]
     num_bounded_loops: int = None  # number of incomplete loops
@@ -485,18 +490,19 @@ def run(
         f"{passfail} {funsig} (paths: {normal}/{len(exs)}, time: {time_info}, bounds: [{', '.join(dyn_param_size)}])"
     )
     for m in models:
-        model, validity, idx, result = m.model, m.validity, m.index, m.result
+        model, is_valid, index, result = m.model, m.is_valid, m.index, m.result
         if result == unsat:
             continue
-        ex = exs[idx]
+        ex = exs[index]
 
-        if model:
-            if validity:
-                print(color_warn(f"Counterexample: {model}"))
+        # model could be an empty dict here
+        if model is not None:
+            if is_valid:
+                print(color_warn(f"Counterexample: {render_model(model)}"))
             elif args.print_potential_counterexample:
                 warn(
                     COUNTEREXAMPLE_INVALID,
-                    f"Counterexample (potentially invalid): {model}",
+                    f"Counterexample (potentially invalid): {render_model(model)}",
                 )
             else:
                 warn(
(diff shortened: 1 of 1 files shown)


karmacoma-eth:
With a simple counterexample:

```
Running 1 tests for test/17_pAtHExPlOSion.t.sol:SyntheticManyBranch
[time] setup: 0.01s (decode: 0.01s, run: 0.00s)
[FAIL] test_pAtHExPlOSion_muly(uint256) (paths: 1023/1025, time: 10.60s (paths: 10.60s, models: 0.00s), bounds: [])
Counterexample: 
    p_x_uint256 = 0x10101000100010000
Symbolic test result: 0 passed; 1 failed; time: 10.93s

[time] total: 12.14s (build: 1.22s, tests: 10.93s)
```

```
{
  "exitcode": 1,
  "test_results": {
    "test/17_pAtHExPlOSion.t.sol:SyntheticManyBranch": [
      {
        "name": "test_pAtHExPlOSion_muly(uint256)",
        "exitcode": 1,
        "num_models": 1,
        "models": [
          {
            "p_x_uint256": "0x10101000100010000"
          }
        ],
        "num_paths": [
          1025,
          1023,
          0
        ],
        "time": [
          10.603914821054786,
          10.600096731912345,
          0.0038180891424417496
        ],
        "num_bounded_loops": 0
      }
    ]
  }
}

```


With an empty counterexample:

```
Running 1 tests for test/18_emptyCounterExample.t.sol:Test18
[FAIL] test_emptyCounterExample() (paths: 0/2, time: 0.01s, bounds: [])
Counterexample: ∅
Symbolic test result: 0 passed; 1 failed; time: 0.02s
```

```
{
  "exitcode": 1,
  "test_results": {
    "test/18_emptyCounterExample.t.sol:Test18": [
      {
        "name": "test_emptyCounterExample()",
        "exitcode": 1,
        "num_models": 1,
        "models": [
          {}
        ],
        "num_paths": null,
        "time": null,
        "num_bounded_loops": null
      }
    ]
  }
}
```

With a ton of counterexamples:

```
{
    "exitcode": 1,
    "test_results": {
        "test/17_pAtHExPlOSion.t.sol:SyntheticManyBranch": [
            {
                "name": "test_pAtHExPlOSion_addy(uint256)",
                "exitcode": 1,
                "num_models": 404,
                "models": [
                    {
                        "p_x_uint256": "0x1000000000000000000"
                    },
                    {
                        "p_x_uint256": "0x100000000000000"
                    },
                    {
                        "p_x_uint256": "0x1000100000000000000"
                    },
                   ...
```

src/halmos/__main__.py
```diff
@@ -820,37 +829,44 @@ def package_result(
...
-def str_model(model, args: Namespace) -> str:
+def to_str_model(model: Model, print_full_model: bool) -> StrModel:
     def select(var):
         name = str(var)
         return name.startswith("p_") or name.startswith("halmos_")
 
-    select_model = filter(select, model) if not args.print_full_model else model
-    formatted = [f"\n    {decl} = {hexify(model[decl])}" for decl in select_model]
+    select_model = filter(select, model) if not print_full_model else model
+    return {str(decl): hexify(model[decl]) for decl in select_model}
```
daejunpark, line 862: hexify is not needed here because it will be done in render_model(), right?

src/halmos/__main__.py
```diff
@@ -820,37 +829,44 @@ def package_result(
...
+
+
+def render_model(model: UnionType[str, StrModel]) -> str:
+    if isinstance(model, str):
+        return model
+
+    formatted = [f"\n    {decl} = {hexify(val)}" for decl, val in model.items()]
```
daejunpark, line 869: `val` is already hexified when generated by to_str_model(), so hexify doesn't seem to be needed here.

daejunpark (approved):

src/halmos/__main__.py
```diff
@@ -537,18 +543,21 @@ def run(
 
     # return test result
     exitcode = 0 if passed else 1
-    num_counterexamples = sum(m.result == sat for m in models)
+    sat_models = [m.model for m in models if m.result == sat]
```
daejunpark, line 546: this includes potentially invalid models even if print_potential_counterexample is not set true.

src/halmos/__main__.py
karmacoma-eth: good catch!

src/halmos/__main__.py
karmacoma-eth: let's keep it in here so that we have hexified output in the json

src/halmos/__main__.py
karmacoma-eth: is that a problem for json output?

karmacoma-eth: yeah it is, fixed in b5edbd1

PR #174, opened 2023-08-15
fix: enable SDIV when --smt-div is on

karmacoma-eth:


daejunpark (approved):

daejunpark:
ah, i didn't look at the expected json change.  the expected json shouldn't be an extended version, as models are non-deterministic and we don't want to see the non-meaningful changes to the expected file in the future prs.

karmacoma-eth:
> the expected json shouldn't be an extended version, as models are non-deterministic and we don't want to see the non-meaningful changes to the expected file in the future prs.

the actual models are ignored in the test output though right? we just check the exitcode and num_models are consistent

daejunpark:
yes, my point was that we don't want to have the counterexamples to be version-controlled, as they are non-deterministic and will keep changing whenever they are re-generated. 

i plan to implement an option that can be used for generating the expected files:
https://github.com/a16z/halmos/issues/175

PR #176, opened 2023-08-15
feat: extended json output by default

### LINKED ISSUE #175: default extended json output
--extended-json-output is more useful in most cases.  make it default, and have a separate option that prints a minimal json output.

daejunpark:
fix #175

new behavior:
- produce extended json output by default
- remove --extended-json-output
- add --minimal-json-output (which generates the previous default output)

test:
- remove counterexamples in the expected json output (this is desired because counterexamples are nondeterministic)

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -546,7 +546,9 @@ def run(
 
     # return test result
     exitcode = 0 if passed else 1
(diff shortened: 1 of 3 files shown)


daejunpark:
this is not ideal, but we can go with this, until #177 is implemented.

PR #183, opened 2023-08-17
feat: summarized warnings

daejunpark:
print a summary of unknown call warnings at the end, instead of printing them on the fly.


karmacoma-eth (approved):
👍

PR #194, opened 2023-09-15
mostly cosmetic changes wrt. storage layout handling

karmacoma-eth:
- `--custom-storage-layout` -> `--storage-layout=generic` (I think it's a bit more intuitive and future proof if we want to add more models in the future)
- rename CustomStorage to GenericStorage
- set the storage model at SEVM creation time, rather than on every store/load


src/halmos/sevm.py
```diff
@@ -1493,18 +1510,13 @@ def arith2(self, ex: Exec, op: int, w1: Word, w2: Word, w3: Word) -> Word:
             raise ValueError(op)
 
     def sload(self, ex: Exec, addr: Any, loc: Word) -> Word:
-        if self.options["custom_storage_layout"]:
-            return CustomStorage.load(ex, addr, loc)
-        else:
-            return SolidityStorage.load(ex, addr, loc)
```
daejunpark, line 1499: oh, is it that a new class instance is created whenever a class method is called?

daejunpark (approved):
looks great, thank you!

src/halmos/sevm.py
karmacoma-eth: no, it just passes the class reference. Here is my reasoning for moving this logic to the SEVM constructor:

- it is a tiny bit more efficient to only check the `"custom_storage_layout"` at config time (rather than every time we process an sload/sstore)
- it reduces code duplication a wee bit
- it feels more conceptually clean: you configure the SEVM class with a specific storage model, and then that SEVM instance will always work with that storage model

PR #198, opened 2023-09-25
test: add nonlinear tests

daejunpark:
add nonlinear benchmark tests

update foundry.toml to avoid unnecessary recompilation


karmacoma-eth (approved):

karmacoma-eth:
do we want to have them under `examples/` rather than `tests/`?

daejunpark:
ah, that might be better!

PR #203, opened 2023-10-11
feat: smt solving refinement

daejunpark:
implement refinement-based non-linear smt solving: optimistically use uninterpreted abstraction for division, and refine the query to use actual division if invalid counterexamples are generated.

remove the following options that are no longer necessary:
- `--smt-div` (as well as other `--smt-*` options)
- `--print-potential-counterexample`

refactor the tests, separating solver tests to be run only on macos runners. (z3 segfaulted for nonlinear queries on ubuntu/windows runners.)

add `--solver-max-memory` option.  see #204.


daejunpark:
karma, does this look good to you? https://github.com/a16z/halmos/pull/203/commits/236378d9988db499993c448dc3c7a641aa6fcbb2

karmacoma-eth:
> karma, does this look good to you? [236378d](https://github.com/a16z/halmos/commit/236378d9988db499993c448dc3c7a641aa6fcbb2)

yes! weird that it worked with the missing `pass`

src/halmos/__main__.py
```diff
@@ -1077,13 +1080,6 @@ def mk_options(args: Namespace) -> Dict:
         "verbose": args.verbose,
         "debug": args.debug,
         "log": args.log,
-        "add": not args.no_smt_add,
-        "sub": not args.no_smt_sub,
-        "mul": not args.no_smt_mul,
-        "div": args.smt_div,
-        "mod": args.smt_mod,
-        "divByConst": args.smt_div_by_const,
-        "modByConst": args.smt_mod_by_const,
```
karmacoma-eth, line 1086: very nice 👌

src/halmos/__main__.py
```diff
@@ -963,38 +966,38 @@ def is_unknown(result: CheckSatResult, model: Model) -> bool:
...
-        sol2.from_string(ex.solver.to_smt2())
-        res = sol2.check()
-        if res == sat:
-            model = sol2.model()
-
-    if is_unknown(res, model) and args.solver_subprocess:
+            print(f"  Checking again with refinement")
+        res, model = solve(refine(ex.solver.to_smt2()), args)
+
+    if args.solver_subprocess and is_unknown(res, model):
```
karmacoma-eth, line 994: does `--solver-subprocess` still serve a purpose here? it may return a sat/unsat even if we got unknown previously because it runs without a timeout?

karmacoma-eth (approved):
LGTM! love the simpler options and handling of invalid counterexamples

src/halmos/__main__.py
daejunpark: yes, and you can run a different solver using this option too.

Issue #200, opened 2023-10-06
Pytest profiles


daejunpark:
Create pytest profiles that can be run both in the ci and locally, given that the pytest configurations are becoming more complicated.

PR #206, opened 2023-10-12
chore: update pyproject.toml

daejunpark:


pyproject.toml
```diff
@@ -37,4 +37,4 @@ target-versions = ["py38", "py39", "py310", "py311"]
 
 [tool.pytest.ini_options]
 # TODO: re-add test_traces.py when we have a better way to support it in CI
-addopts = "--ignore=tests/lib/solady/ext/woke --ignore=tests/test_traces.py"
```
karmacoma-eth, line 40: I run into problems running pytest at the root because of this lib

karmacoma-eth (approved):

pyproject.toml
daejunpark: yeah, #200 will be a nicer and longer term solution, but for now, we can go with this.

btw, i've made the change to ignore the entire tests/lib, because we may add other libraries there later that have their own pytest which we don't want to run.

PR #209, opened 2023-10-18
feat: nicer output

karmacoma-eth:
Includes a bunch of things:
- type based counterexample rendering (different rendering for uints, signed ints, address, bool, bytes, string, bytes32...)
- console.log functions leverage the same type based rendering
- includes a bunch of new console.log functions + tests
- updates to latest halmos-cheatcode, implements the new cheatcodes (createInt, createInt256, createBytes4, createString) and add tests for these functions

Plus some internal reorganizing:
- move utils-like functions from sevm.py to utils.py
- move cheatcode and console.log handling logic from sevm.py to cheatcodes.py
- switch to a lookup+handler based mechanism in console and halmos_cheat_code

### DIFF
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -179,7 +179,7 @@ class halmos_cheat_code:
             raise HalmosException(f"bitsize larger than 256: {bits}")
 
         name = name_of(extract_string_argument(arg, 1))
-        return halmos_cheat_code.create_generic(ex, bits, name, f"int{bits}")
+        return int256(halmos_cheat_code.create_generic(ex, bits, name, f"int{bits}"))
 
     @staticmethod
     def create_int256(ex, arg):
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -71,11 +71,23 @@ class halmos_cheat_code:
     # bytes4(keccak256("createUint(uint256,string)"))
     create_uint: int = 0x66830DFA
 
+    # bytes4(keccak256("createUint256(string)"))
+    create_uint256: int = 0xBC7BEEFC
+
+    # bytes4(keccak256("createInt(uint256,string)"))
+    create_int: int = 0x49B9C7D4
+
+    # bytes4(keccak256("createInt256(string)"))
+    create_int256: int = 0xC2CE6AED
+
     # bytes4(keccak256("createBytes(uint256,string)"))
     create_bytes: int = 0xEEF5311D
 
-    # bytes4(keccak256("createUint256(string)"))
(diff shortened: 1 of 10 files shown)


src/halmos/utils.py
```diff
@@ -36,13 +213,80 @@ def hexify(x):
...
+    elif type_name == "string":
+        return render_string(val)
+    elif type_name == "bytes":
+        return render_bytes(val)
+    elif type_name == "address":
+        return render_address(val)
+    else:  # bytes32, bytes4, structs, etc.
+        return hexify(val)
+
+
```
daejunpark, line 289: i think we need to catch potential exceptions raised by the render functions here.  even if `is_bv_value(val)` is true, the render functions could still raise an exception, when the type of `val` somehow differs from expected, e.g., `.decode("utf-8")` may fail in `render_string()`, or `b.hex()` may fail if `b` is not of type `bytes`.

daejunpark (commented):
looks great!  the refactoring is amazing!  thank you so much!

my only comment is about handling potential exceptions in stringify.

src/halmos/cheatcodes.py
```diff
@@ -154,11 +318,262 @@ class hevm_cheat_code:
...
+        0x2D5B6CB9: log_int256,
+        0xB60E72CC: log_string_uint256,
+    }
+
+    @staticmethod
+    def handle(ex, arg: BitVec) -> None:
+        funsig: int = int_of(extract_funsig(arg), "symbolic console function selector")
+
+        if handler := console.handlers.get(funsig):
+            return handler(arg)
```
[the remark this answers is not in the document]
karmacoma-eth: you're right, added a test for it in 8b5b151:

```
Running 1 tests for test/Console.t.sol:ConsoleTest
console.handle: UnicodeDecodeError('utf-8', b'\xff', 0, 1, 'invalid start byte') with arg=0x4b5c4277000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000002c74686973206973206120737472696e67207468617420776f6e2774206465636f646520746f207574662d383a00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001ff00000000000000000000000000000000000000000000000000000000000000
[PASS] check_log_undecodable_string() (paths: 1/1, time: 0.05s, bounds: [])
Symbolic test result: 1 passed; 0 failed; time: 0.05s
```

src/halmos/utils.py
karmacoma-eth: in 5b3c56f

daejunpark (approved):

PR #211, opened 2023-10-25
Update __main__.py

aviggiano:
Update `args.function` to filter by regular expression instead of string prefix.

This allows us to target specific functions by passing, e.g., `halmos --function '\btest_add_maximum_value\b'`


daejunpark:
@aviggiano thank you so much! This has been extended to #238.

PR #218, opened 2023-11-10
perf: single solver shared by multiple paths

daejunpark:
- share a single solver for all paths
- full dfs path exploration across external call boundaries
- use z3 push/pop to avoid the solver reinstantiation overhead during the dfs
- generate counterexamples in parallel with the dfs
- a new option for early exit upon finding the first counterexample
- a new option to set the number of threads for parallel solvers

note: the farcaster tests are temporarily disabled in ci, due to independent issues that are to be fixed in a separate pr


src/halmos/sevm.py
```diff
@@ -815,7 +864,7 @@ def sha3_data(self, data: Bytes, size: int) -> Word:
         sha3_expr = f_sha3(data)
 
         # assume hash values are sufficiently smaller than the uint max
-        self.solver.add(ULE(sha3_expr, con(2**256 - 2**64)))
+        self.path.append(ULE(sha3_expr, con(2**256 - 2**64)))
```
karmacoma-eth, line 867: do you think it would be useful to expose an additional method like `add_assertion`? It feels like this is a way to add it to the solver assertions but as a side effect it is also appended to the list of human-readable path conditions, when it doesn't really correspond to a path (branching) condition

karmacoma-eth: or just `path.assert(cond)` for short

src/halmos/sevm.py
```diff
@@ -596,6 +643,7 @@ class Exec:  # an execution path
 
     # tx
     context: CallContext
+    callback: Any  # to be called when returning back to the parent context
```
karmacoma-eth, line 646: can we use something like `Optional[Callable]` as the type hint here?

src/halmos/sevm.py
```diff
@@ -625,7 +672,9 @@ def __init__(self, **kwargs) -> None:
         self.balance = kwargs["balance"]
         #
         self.block = kwargs["block"]
+        #
         self.context = kwargs["context"]
+        self.callback = kwargs["callback"]
```
karmacoma-eth, line 677: given that callback is optional (can be None), can we also make it optional in the kwargs?

something like `self.callback = kwargs.get("callback", None)`

src/halmos/sevm.py
```diff
@@ -2049,38 +2106,37 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps, HalmosLogs]:
...
-                        steps[step_id] = {"parent": prev_step_id, "exec": str(ex)}
+                        self.steps[step_id] = {"parent": prev_step_id, "exec": str(ex)}
                     # elif opcode == EVM.CALL:
-                    #     steps[step_id] = {'parent': prev_step_id, 'exec': str(ex) + ex.st.str_memory() + '\n'}
+                    #     self.steps[step_id] = {'parent': prev_step_id, 'exec': str(ex) + ex.st.str_memory() + '\n'}
                     else:
-                        # steps[step_id] = {'parent': prev_step_id, 'exec': ex.summary()}
-                        steps[step_id] = {"parent": prev_step_id, "exec": str(ex)}
+                        # self.steps[step_id] = {'parent': prev_step_id, 'exec': ex.summary()}
+                        self.steps[step_id] = {"parent": prev_step_id, "exec": str(ex)}
```
karmacoma-eth, line 2114: not directly related to this PR, but it looks like both branches are doing the same thing and could be simplified

src/halmos/sevm.py
```diff
@@ -2394,17 +2450,26 @@ def run(self, ex0: Exec) -> Tuple[List[Exec], Steps, HalmosLogs]:
...
+
                 ex.halt(data=None, error=err)
-                out.append(ex)
-                continue
 
-        return (out, steps, logs)
+                if ex.callback is None:
+                    yield ex
+                else:
+                    yield from ex.callback(ex, stack, step_id)
```
karmacoma-eth, line 2470: can we extract these 4 lines in a helper function? They seem to be duplicated a couple times

a comment might help understanding the `yield from ex.callback(...)`, it takes some thinking to figure out 🧙‍♂️

src/halmos/sevm.py
```diff
@@ -585,6 +585,53 @@ def __next__(self) -> Instruction:
...
+        self.solver = solver
+        self.num_scopes = 0
+        self.conditions = []
+        self.pending = []
+
+    def __deepcopy__(self, memo):
+        if len(self.pending) > 0:
+            raise ValueError("deepcopy pending path", self)
+
+        path = Path(self.solver)
```
karmacoma-eth, line 604: might be worth a comment here. This is where the real magic happens since we re-use the solver instance instead of actually copying it, right? Maybe explain in a sentence or two why we do this and the trade-offs

src/halmos/sevm.py
```diff
@@ -1960,20 +2019,21 @@ def jump(self, ex: Exec, stack: List[Tuple[Exec, int]], step_id: int) -> None:
...
     def create_branch(self, ex: Exec, cond: BitVecRef, target: int) -> Exec:
-        new_solver = create_solver(
-            timeout=self.options["timeout"], max_memory=self.options["max_memory"]
-        )
-        new_solver.add(ex.solver.assertions())
-        new_solver.add(cond)
-        new_path = ex.path.copy()
-        new_path.append(str(cond))
+        new_path = deepcopy(ex.path)
+        new_path.pending.append(cond)
```
karmacoma-eth, line 2032: style: I had to hunt for this, I couldn't immediately see in Path where things were added to `pending`. Would be mildly cleaner/easier with an `add_pending(cond)` method IMO

src/halmos/sevm.py
```diff
@@ -585,6 +585,53 @@ def __next__(self) -> Instruction:
         return insn
 
 
+class Path:
```
karmacoma-eth, line 588: nit: would you say that an instance of this class does represent a single path through the program? My instinct says no, because it makes it a little weird to initialize an SEVM with a Path object (since I don't really have a path before we get started). 

Semantically, would something like `PathTracker` or something similar be more accurate?

karmacoma-eth (approved):
Very clean! See minor comments and requests for clarifications

src/halmos/sevm.py
daejunpark: fair point!  but in my mental model, it's still a path.  i added explanation in the comment for now.  we can discuss more to improve naming.

src/halmos/sevm.py
daejunpark: the callback is None only at the top-level.  i prefer to non-optional in the kwargs, so that i can catch earlier if callback is missing.

src/halmos/sevm.py
daejunpark: good point!  added a todo note to get back to this later.

src/halmos/sevm.py
daejunpark: good catch! it's legacy code, and i need to clean this up anyway; will do this later.

PR #231, opened 2023-12-20
fix: trace generation for shared solver

daejunpark:
#218 introduced new issues for trace generation:  rendering z3 objects might fail when the shared solver is used for another path context.

fix the issues by:
- generate traces in advance at the end of each path before switching to another path
- but generate traces only when the trace rendering flags are enabled
- maintain branching conditions separately, so that internal constraints don't appear in trace outputs

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -256,7 +256,9 @@ def render_trace(context: CallContext, file=sys.stdout) -> None:
         # TODO: select verbosity level to render full initcode
         # initcode_str = rendered_initcode(context)
         initcode_str = f"<{byte_length(message.data)} bytes of initcode>"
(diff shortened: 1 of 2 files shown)


src/halmos/__main__.py
```diff
@@ -228,7 +229,15 @@ def rendered_log(log: EventLog) -> str:
     return f"{opcode_str}({args_str})"
 
 
-def render_trace(context: CallContext) -> None:
+def rendered_trace(context: CallContext) -> str:
+    output = io.StringIO()
+    render_trace(context, file=output)
+    result = output.getvalue()
+    output.close()
+    return result
```
karmacoma-eth, line 237: pretty neat! Suggest using a context manager here to make it even crispier

```suggestion
def rendered_trace(context: CallContext) -> str:
    with io.StringIO() as output:
        render_trace(context, file=output)
        return output.getvalue()
```

PR #238, opened 2023-12-21
feat: add --match-test and --match-contract

daejunpark:
add --match-test REGEX:
- the `--function PREFIX` is now a shortcut for `--match-test ^{PREFIX}`.
- unless REGEX starts with `^`, the prefix is automatically added, i.e., the final regex is `^{PREFIX}.*{REGEX}`.

add --match-contract REGEX:
- the `--contract NAME` is now a shortcut for `--match-contract ^{NAME}$`.
- ignored if `--contract` is given.

[subsume #211]


karmacoma-eth (approved):
nice, ship it👌

PR #239, opened 2023-12-22
paper cut fixes

karmacoma-eth:
- close https://github.com/a16z/halmos/issues/220 (as a band aid until we have a cleaner solution)
- allow console.log to render symbolic values

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -246,10 +246,6 @@ def wstore_partial(
         raise ValueError(data)
 
 
-def is_byte(x: Any) -> bool:
-    return x in range(0, 256) or (is_bv(x) and eq(x.sort(), BitVecSort8))
-
-
 def wstore_bytes(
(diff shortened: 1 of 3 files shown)


src/halmos/sevm.py
```diff
@@ -253,9 +253,10 @@ def wstore_bytes(
         raise ValueError(size, arr)
     wextend(mem, loc, size)
     for i in range(size):
-        if not eq(arr[i].sort(), BitVecSort8):
+        arr_i_bv = arr[i] if is_bv(arr[i]) else con(arr[i], 8)
```
daejunpark, line 256: questions:  when arr[i] is not bv:
- do we want to store them as bv, rather than keeping it int?
- do we need to check if arr[i] < 256, or is it guaranteed?

karmacoma-eth: nice catch, can you check again with [7c5c364](https://github.com/a16z/halmos/pull/239/commits/7c5c3647f8c4a99451a1852ce607977d9098d5fe)?

PR #242, opened 2023-12-23
calldata rendering bugfix, large memory offset bugfix

karmacoma-eth:


src/halmos/__main__.py
```diff
@@ -237,6 +237,15 @@ def rendered_trace(context: CallContext) -> str:
         return output.getvalue()
 
 
+def rendered_calldata(calldata: List[Byte]) -> str:
+    if any(is_bv(x) for x in calldata):
+        # make sure every byte is wrapped
+        calldata_bv = [x if is_bv(x) else con(x, 8) for x in calldata]
+        return hexify(simplify(Concat(calldata_bv)))
```
daejunpark, line 244: `Concat` will fail if the given list is a singleton. `concat` should be used.

karmacoma-eth: Concat, why are you so finicky? 🙈

karmacoma-eth: in c430f4e

src/halmos/sevm.py
```diff
@@ -39,6 +39,8 @@
 EMPTY_BYTES = b""
 MAX_CALL_DEPTH = 1024
 
+# TODO: make this configurable
+MAX_MEMORY_SIZE = 2**20
```
daejunpark, line 43: can this be increased to 2^32 too?

daejunpark (approved):

PR #249, opened 2024-01-23
Small bytecode caching optimizations

karmacoma-eth:
- cache bytecode length
- cache instruction access: this one is a memory-compute tradeoff, but I expect it to be favorable most of the time because program bytecode is small but accessed frequently. Can do a lru_cache if memory consumption turns out to be a problem
- separate __getitem__ and .slice so that __getitem__ can be cached
- return cached jump destinations instead of redecoding on the fly in jumpi_id

All together, improve the runtime of some tests by about 5%

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -6,7 +6,7 @@ import re
 from copy import deepcopy
 from collections import defaultdict
 from dataclasses import dataclass, field
-from functools import cache, reduce
+from functools import reduce, lru_cache
 from typing import (
     Any,
     Callable,
@@ -591,7 +591,7 @@ class Contract:
         data = padded_slice(self._rawcode, start, size, default=0)
         return bytes(data)
 
-    @cache
+    @lru_cache(maxsize=256)
     def __getitem__(self, key: int) -> UnionType[int, BitVecRef]:
         """Returns the byte at the given offset."""
         offset = int_of(key, "symbolic index into contract bytecode {offset!r}")
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -1,13 +1,12 @@
 # SPDX-License-Identifier: AGPL-3.0
 
-import json
 import math
 import re
 
 from copy import deepcopy
 from collections import defaultdict
 from dataclasses import dataclass, field
-from functools import reduce
+from functools import cache, reduce
 from typing import (
     Any,
     Callable,
@@ -533,6 +532,9 @@ class Contract:
 
         self._rawcode = rawcode
 
+        # cache length since it doesn't change after initialization
+        self._length = byte_length(rawcode)
+
     def __init_jumpdests(self):
         self.jumpdests = set()
 
@@ -563,7 +565,7 @@ class Contract:
         opcode = int_of(self[pc])
 
         if EVM.PUSH1 <= opcode <= EVM.PUSH32:
-            operand = self[pc + 1 : pc + opcode - EVM.PUSH0 + 1]
+            operand = self.slice(pc + 1, pc + opcode - EVM.PUSH0 + 1)
             return Instruction(opcode, pc=pc, operand=operand)
 
         return Instruction(opcode, pc=pc)
@@ -572,16 +574,10 @@ class Contract:
         opcode = self[pc]
         return pc + instruction_length(opcode)
 
-    def __getslice__(self, slice):
-        step = 1 if slice.step is None else slice.step
-        if step != 1:
-            return ValueError(f"slice step must be 1 but got {slice}")
-
+    def slice(self, start, stop) -> UnionType[bytes, BitVecRef]:
         # symbolic
         if is_bv(self._rawcode):
-            extracted = extract_bytes(
-                self._rawcode, slice.start, slice.stop - slice.start
-            )
+            extracted = extract_bytes(self._rawcode, start, stop - start)
 
             # check if that part of the code is concrete
             if is_bv_value(extracted):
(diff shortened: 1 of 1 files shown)


src/halmos/sevm.py
```diff
@@ -646,6 +640,9 @@ def __next__(self) -> Instruction:
         if self.pc >= len(self.contract):
             raise StopIteration
 
+        if self.is_symbolic and is_bv(self.contract[self.pc]):
+            raise StopIteration
+
```
daejunpark, line 645: i couldn't understand this.  why stop iterating at a symbolic byte?

src/halmos/sevm.py
```diff
@@ -1047,8 +1044,7 @@ def is_jumpdest(self, x: Word) -> bool:
         if pc < 0:
             raise ValueError(pc)
 
-        opcode = unbox_int(self.pgm[pc])
-        return opcode == EVM.JUMPDEST
+        return pc in self.pgm.valid_jump_destinations()
```
daejunpark, line 1047: i don't think it is correct.  because of the stopping behavior of CodeIterator, `valid_jump_destinations()` may not include all jump destinations, i.e., the ones after a symbolic byte.  e.g., immutable storage variables can be initialized with a symbolic value, then the symbol appear in the bytecode, and then CodeIterator stops at the immutable symbol, so any jumpdest opcode following the symbol will not be included in the valid_jump_destinations set.

daejunpark:
5% speedup sounds good, but i'd like to know how much memory overhead this would incur.  (we had hard time with reducing memory recently.) to better evaluate the overall performance benefit, could you please provide the performance comparison data (for both time and memory) for all benchmark tests?

src/halmos/sevm.py
karmacoma-eth: in this case we assume that symbolic arguments are data and don't contain potentially valid jumpdests (technically they could, but we don't normally attempt to deal with symbolic bytecode)

daejunpark: in the runtime bytecode, the immutable variables appear in the middle of code (as push opcode arguments), not at the end of the code. if the immutable variables are initialized with symbolic values, then we have symbolic values in the middle of code, and we shouldn't stop there.

karmacoma-eth:
> 5% speedup sounds good, but i'd like to know how much memory overhead this would incur. (we had hard time with reducing memory recently.) to better evaluate the overall performance benefit, could you please provide the performance comparison data (for both time and memory) for all benchmark tests?

profiling memory is tricky, let's do a quick synthetic test:

```
def test_contract_memoization_memory():
    import resource

    code = Contract("5b" * 24576)

    maxrss_before = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print("max memory usage before iteration:", maxrss_before)
    for insn in code:
        assert insn.opcode == 0x5b

    maxrss_after1 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print(f"after iteration: +{(maxrss_after1 - maxrss_before) / 1024**2:.2f}MB")
    # after iteration: +4.22MB

    for i in range(len(code)):
        assert code[con(i)] == 0x5b

    maxrss_after2 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print(f"after bv_value lookups: +{(maxrss_after2 - maxrss_after1) / 1024**2:.2f}MB")
    # after bv_value lookups: +62.79MB
```

the problem is that z3 objects can get cached too, so the memory usage can get much worse than with concrete offsets

let me do a quick test with an lru_cache

karmacoma-eth:
> > 5% speedup sounds good, but i'd like to know how much memory overhead this would incur. (we had hard time with reducing memory recently.) to better evaluate the overall performance benefit, could you please provide the performance comparison data (for both time and memory) for all benchmark tests?
> 
> profiling memory is tricky, let's do a quick synthetic test:
> 
> ```python
> def test_contract_memoization_memory():
>     import resource
> 
>     code = Contract("5b" * 24576)
> 
>     maxrss_before = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
>     print("max memory usage before iteration:", maxrss_before)
>     for insn in code:
>         assert insn.opcode == 0x5b
> 
>     maxrss_after1 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
>     print(f"after iteration: +{(maxrss_after1 - maxrss_before) / 1024**2:.2f}MB")
>     # after iteration: +4.22MB
> 
>     for i in range(len(code)):
>         assert code[con(i)] == 0x5b
> 
>     maxrss_after2 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
>     print(f"after bv_value lookups: +{(maxrss_after2 - maxrss_after1) / 1024**2:.2f}MB")
>     # after bv_value lookups: +62.79MB
> ```
> 
> the problem is that z3 objects can get cached too, so the memory usage can get much worse than with concrete offsets
> 
> let me do a quick test with an lru_cache

with `@lru_cache(maxsize=256)` (so 100x less caching), we get:

```
after iteration: +0.14MB
after bv_value lookups: +0.33MB
```

seems like a better trade off, wdyt?

src/halmos/sevm.py
karmacoma-eth: discussed in person: I believe this is correctly handled as long as symbolic bytes are instruction arguments

daejunpark: now i also see that it cannot iterate beyond this point, as the argument size of a symbolic opcode cannot be determined.

src/halmos/sevm.py
daejunpark: i was wrong. the iterator iterates over instructions (that consist of opcode and arguments if any), not over bytes. since the symbolic immutable values appear as an argument of the push opcode, the iterator will not stop prematurely there.

but there is still a case where this change makes differences. if there is a symbolic opcode in the middle of bytecode that is not reachable (deadcode), then the iterator approach could miss the reachable part of bytecode after the deadcode. this seems to be a fundamental question how to precisely compute valid jumpdests in the presence of possibly symbolic deadcode. since this is an orthogonal topic, and it is not likely to happen in most practical cases, we can merge this pr for now, and have a separate discussion (probably in #251 and #252).

daejunpark (approved):
thanks for this low-level optimization!  one step at a time, and eventually these efforts will culminate in significant improvement!

src/halmos/sevm.py
karmacoma-eth: true, in all the cases I considered the data-in-bytecode was appended at the end:

- metadata
- constructor arguments in init code
- jump table values

So these would not cause an issue, but in principle it's certainly possible to have a jump table at the beginning of bytecode. Maybe to handle this we could consider that symbolic values in bytecode implicitly are data, and we could resume jumpdest processing when concrete data resumes?

daejunpark: That would be an option, but if the data section consists of both symbolic and concrete values, then it wouldn't be trivial to determine when resuming jumpdest processing. Perhaps we can move this discussion to #252 ?

PR #250, opened 2024-01-25
Dump refined queries

karmacoma-eth:
Problem: previously, we would only dump unrefined queries.

By moving the dumping inside solve, we dump every assertion query, including refined queries.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -335,7 +335,7 @@ def run_bytecode(hexcode: str, args: Namespace) -> List[Exec]:
         else:
             print(f"Final opcode: {mnemonic(opcode)})")
             print(f"Return data: {returndata}")
-            dump_dirname = f"/tmp/halmos-{uuid.uuid4()}"
+            dump_dirname = f"/tmp/halmos-{uuid.uuid4().hex}"
             model_with_context = gen_model_from_sexpr(
                 GenModelArgs(args, idx, ex.path.solver.to_smt2(), dump_dirname)
             )
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -1021,6 +1021,7 @@ def solve(
         with open(dump_filename, "w") as f:
             f.write("(set-logic QF_AUFBV)\n")
             f.write(query)
+            f.write("(get-model)\n")
 
     solver = mk_solver(args, ctx=Context(), assertion=True)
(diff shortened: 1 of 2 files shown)


src/halmos/__main__.py
```diff
@@ -1015,13 +1007,22 @@ class GenModelArgs:
...
 
-def solve(query: str, args: Namespace) -> Tuple[CheckSatResult, Model]:
+def solve(
+    query: str, args: Namespace, dump_filename: Optional[str] = None
+) -> Tuple[CheckSatResult, Model]:
+    if dump_filename:
+        with open(dump_filename, "w") as f:
+            f.write("(set-logic QF_AUFBV)\n")
+            f.write(query)
+            f.write("(get-model)\n")
```
daejunpark, line 1024: `(get-model)` will be an error if the result is unsat.  indeed, z3 provides a command line option that automatically prints a model if sat.  if other smt solvers have such an option, we wouldn't need this.

src/halmos/__main__.py
```diff
@@ -335,8 +335,9 @@ def run_bytecode(hexcode: str, args: Namespace) -> List[Exec]:
         else:
             print(f"Final opcode: {mnemonic(opcode)})")
             print(f"Return data: {returndata}")
+            dump_dirname = f"/tmp/halmos-{uuid.uuid4()}"
```
daejunpark, line 338: `uuid.uuid4().hex` to be consistent?

daejunpark, line 338:
```suggestion
            dump_dirname = f"/tmp/halmos-{uuid.uuid4().hex}"
```

src/halmos/__main__.py
daejunpark: turned out that some smt solvers don't provide such an option, so there is no clear solution that works for all solvers.  for now, we can keep this knowing that the query will generate error when unsat, and come back to this later if needed.

daejunpark (approved):

src/halmos/__main__.py
karmacoma-eth: sorry forgot to update this: I checked and all the solvers are tried are ok with `(get-model)` even in the unsat case. 

z3 does prints an error, but that's after correctly returning `unsat` so I think it's good overall. 

One thing I still need to check if `(get-model)` makes a big difference in the `sat` case

Issue #251, opened 2024-01-26
Validate legality of jumpdests


0xkarmacoma:
**Is your feature request related to a problem? Please describe.**

Currently, you can jump anywhere in bytecode (inluding non-jumpdests or the middle of a PUSH), we do not validate that the bytecode jumps on a valid jumpdest

PR #252, opened 2024-01-26
[WIP] Validate jumpdests

karmacoma-eth:
Validate the legality of jump destination targets
    
Previously, you could jump anywhere in bytecode (inluding non-jumpdests or the middle of a PUSH), we did not validate that the bytecode jumped on a valid jumpdest

See #251 

TODO:

- [X] test_illegal_jump
- [X] implement jumpdest validation for JUMP
- [ ] test_illegal_jumpi
- [ ] implement jumpdest validation for JUMPI

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -167,7 +167,7 @@ def mnemonic(opcode) -> str:
 
 
 def instruction_length(opcode: Any) -> int:
-    opcode = int_of(opcode)
+    opcode = int_of(opcode, "can not determine instruction length for symbolic opcode")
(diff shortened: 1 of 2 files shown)


daejunpark:
a question copied from https://github.com/a16z/halmos/pull/249#discussion_r1481988754

if there is a symbolic opcode in the middle of bytecode that is not reachable (deadcode), then the iterator approach could miss the reachable part of bytecode after the deadcode. this seems to be a fundamental question how to precisely compute valid jumpdests in the presence of possibly symbolic deadcode.

karmacoma-eth:
closing for now, hasn't been top priority

PR #255, opened 2024-03-01
Add --build-info to forge build command

karmacoma-eth:
Forge has recently (https://github.com/foundry-rs/foundry/pull/7197) dropped the ast from the build json by default, so we need to request it explicitly with `--ast`, `ast = <bool>` config or just `--build-info` to avoid checking if forge supports the `--ast` command


gakonst:
Sorry about that -- we should have notified users :)

karmacoma-eth:
@gakonst one day I will get the @0xalpharush treatment #lifegoals ;)

PR #258, opened 2024-03-19
Small fixes for `getting-started.md`

cos:


docs/getting-started.md
```diff
@@ -58,8 +58,10 @@ contract MyToken is ERC20 {
 Then you can write a `setUp()` function that creates a new token contract with a _symbolic_ initial supply, as follows:
 ```solidity
+import "forge-std/Test.sol";
+import "./MyToken.sol";
```
daejunpark, line 62: great!  just to better follow the modern foundry style:
```suggestion
import {Test} from "forge-std/Test.sol";
import {MyToken} from "../src/MyToken.sol";
```

docs/getting-started.md
```diff
@@ -104,7 +106,7 @@ Below is an example symbolic test for the token transfer function:
 function check_transfer(address sender, address receiver, uint256 amount) public {
     // specify input conditions
     vm.assume(receiver != address(0));
-    vm.assume(token.balance(sender) >= amount);
+    vm.assume(token.balanceOf(sender) >= amount);
```
daejunpark, line 109: good catch! could you please fix the same mistake below (at line 184)?
```
vm.assume(token.balance(sender) >= amount);
```

daejunpark (commented):
thanks! just minor further changes request!

PR #259, opened 2024-03-23
feat: constrain (r, s, v) signature values

karmacoma-eth:
+ BadElections in examples/simple
+ [test for EIP-2098 compact signatures](https://github.com/a16z/halmos/commit/81cb2ad4cd622636ecc285e229e1bf8f2f55c764)

### DIFF
--- a/src/halmos/utils.py
+++ b/src/halmos/utils.py
@@ -182,8 +182,13 @@ def extract_string_argument(calldata: BitVecRef, arg_idx: int):
     return string_bytes.decode("utf-8") if is_concrete(string_bytes) else string_bytes
 
 
-def extract_bytes(data: BitVecRef, byte_offset: int, size_bytes: int) -> BitVecRef:
+def extract_bytes(
+    data: Optional[BitVecRef], byte_offset: int, size_bytes: int
+) -> BitVecRef:
     """Extract bytes from calldata. Zero-pad if out of bounds."""
+    if data is None:
+        return BitVecVal(0, size_bytes * 8)
+
     n = data.size()
     if n % 8 != 0:
         raise ValueError(n)
--- a/examples/simple/src/BadElections.sol
+++ b/examples/simple/src/BadElections.sol
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity >=0.8.0 <0.9.0;
+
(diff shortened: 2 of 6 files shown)


src/halmos/utils.py
```diff
@@ -182,8 +182,13 @@ def extract_string_argument(calldata: BitVecRef, arg_idx: int):
...
 
 
-def extract_bytes(data: BitVecRef, byte_offset: int, size_bytes: int) -> BitVecRef:
+def extract_bytes(
+    data: Optional[BitVecRef], byte_offset: int, size_bytes: int
+) -> BitVecRef:
     """Extract bytes from calldata. Zero-pad if out of bounds."""
+    if data is None:
+        return BitVecVal(0, size_bytes * 8)
+
```
daejunpark, line 191: could you add a regression test that would fail without this fix? (it's not clear when None can be passed to this function.) also, does this fix the `vm.etch(x, "")` case as well?

karmacoma-eth: it's related but separate from the `vm.etch(x, "")` issue, we don't have a very nice way to deal with empty bytes in general. They get sometimes represented as `bytes()`, sometimes as `[]`, sometimes as None

the test that revealed this issue is `check_ecrecover_invalidCallReturnsNothing`, in the following line:
`let succ := call(gas(), ECRECOVER_PRECOMPILE, 0, 0, 0, 0, 0)`

daejunpark: thanks. now i can see where None comes from. as you mentioned, i agree that we need a better way to handle nicely empty bytes. could you please open an issue about this to discuss further later?

daejunpark (approved):

daejunpark:
@karmacoma-eth wdyt? 
https://github.com/a16z/halmos/commit/30c6bb741d2e6d50136f389eed0b9ac748e9bc58#r141719157

karmacoma-eth:
> @karmacoma-eth wdyt? [30c6bb7#r141719157](https://github.com/a16z/halmos/commit/30c6bb741d2e6d50136f389eed0b9ac748e9bc58#r141719157)

replied to the question on the commit, but copying here just in case:

> I don't think we can do that unfortunately, EIP-2 specifies only the behavior for transaction signatures.

> AFAIK, the ecrecover precompile is still perfectly happy with s values in the full range

PR #263, opened 2024-03-26
test: add expected output for new test

daejunpark:
add missing expected output for the new test added in #259


karmacoma-eth:
ah oops thank you!

karmacoma-eth (approved):

PR #264, opened 2024-03-26
fix: vm.etch(addr, "") used to raise an exception

karmacoma-eth:
this properly handles empty bytecode arguments, and sets the code at the address to the empty sequence


daejunpark (approved):

daejunpark:
just waiting for:
https://github.com/a16z/halmos/actions/runs/8445205851
https://github.com/a16z/halmos/actions/runs/8445206431

PR #266, opened 2024-03-28
docs: add erc4626 test example

daejunpark:
currently erc4626 tests fail due to timeout


daejunpark:
this test currently doesn't work due to the limitation of smt solving. closing this for now, but will keep the branch, as a reference for the erc4626 test examples.

PR #271, opened 2024-04-09
fix: smt encoding for evm div-by-zero

daejunpark:
Fix the SMT encoding of division (and modulo) by zero.

In EVM, division (and modulo) by zero is defined to be zero, whereas they are defined differently in [SMT-LIB](https://smt-lib.org/theories-FixedSizeBitVectors.shtml).

Thus, the EVM division `(evm_bvudiv_N x y)` should be encoded as:
```
(ite (= y (_ bv0 N)) (_ bv0 N) (bvudiv x y))
```
where `N` is the bitsize. The EVM modulo operation should be encoded similarly.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -1205,17 +1205,23 @@ def is_unknown(result: CheckSatResult, model: Model) -> bool:
 
 def refine(query: SMTQuery) -> SMTQuery:
     smtlib = query.smtlib
+
     # replace uninterpreted abstraction with actual symbols for assertion solving
-    # TODO: replace `(f_evm_bvudiv x y)` with `(ite (= y (_ bv0 256)) (_ bv0 256) (bvudiv x y))`
-    #       as bvudiv is undefined when y = 0; also similarly for f_evm_bvurem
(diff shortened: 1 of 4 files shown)


daejunpark:
Can be merged after #340

karmacoma-eth:
it would be nice to have a test that demonstrates the issue that this fixes. What about this?

```solidity
    function check_Div(uint num, uint256 den) public pure {
        vm.assume(num > den);
        uint256 y;

        assembly {
            y := div(num, den)
        }

        assert(y != 0);
    }
```

We currently get

```
# of potential paths involving assertion violations: 1 / 2  (--solver-threads 20)
  Checking again with refinement
  Invalid path; ignored (path id: 1)
[PASS] check_Div(uint256,uint256) (paths: 2, time: 12.51s (paths: 0.03s, models: 12.49s), bounds: [])
```

but we should get a counterexample with den == 0, right?

daejunpark:
new tests are added: https://github.com/a16z/halmos/pull/271/commits/a5199dfb73404fd2c1095a2107b7f4da04771f7e

karmacoma-eth (approved):
neat, thank you!

PR #274, opened 2024-04-17
Add GitHub Codespaces dev container config

cos:
- Adds a GitHub Codespaces dev container configuration
- Adds `pytest` to `requirements-dev.txt`

Tested by executing the following in a fresh dev container:
```
source .venv/bin/activate
python -m pytest -v -k "not long and not ffi" --ignore=tests/lib --halmos-options="-v -st --storage-layout solidity --solver-timeout-assertion 0"
```

# GitHub Codespaces tab

Screenshot 2024-04-16 at 6 56 28 PM


.devcontainer/postCreate.sh
```diff
@@ -0,0 +1,22 @@
...
+python -m pip install -r requirements.txt
+python -m pip install -r requirements-dev.txt
+pre-commit install
+pre-commit run --all-files
+python -m pip install -e .
+
+# Installing during postCreateCommand instead of at docker creation
+# because foundryup installs the binaries in place not accessible by the
+# current (vscode) user.
+deactivate
```
karmacoma-eth, line 13: Is this needed? I wouldn't expect the venv to interfere with the foundry install

.devcontainer/postCreate.sh
```diff
@@ -0,0 +1,22 @@
...
+mkdir foundry
+cd foundry
```
karmacoma-eth, line 15: I don't think we need these, no files are created in this foundry dir right?

CONTRIBUTING.md
```diff
@@ -51,6 +51,15 @@ python -m black .
 
 [black]: <https://black.readthedocs.io/en/stable/>
 
+## GitHub Codespace
+
+A pre-configured development environment is available as a GitHub Codespaces dev container.
+
+After starting the container, make sure to activate the virtual environment:
+```
+source .venv/bin/activate
```
karmacoma-eth, line 60: out of curiosity, would this be needed if we didn't have `deactivate` in the post create command? (I'm not very familiar with codespaces)

requirements-dev.txt
```diff
@@ -1,2 +1,3 @@
 black
 pre-commit
+pytest
```
karmacoma-eth, line 3: good catch, thank you

.devcontainer/postCreate.sh
cos: iirc,  foundryup changes PATH, which is tied to the python env

cos: hm, looks like it works without, removed

.devcontainer/postCreate.sh
cos: yep, leftover from a different way of doing this, removed

CONTRIBUTING.md
cos: yes, post create runs in a different terminal -- the standard way of doing this [is complicated](https://docs.github.com/en/codespaces/setting-your-user-preferences/personalizing-github-codespaces-for-your-account#dotfiles), but lemme try something, might find a way around

karmacoma-eth (approved):
very nice, thank you for the updates @cos

PR #284, opened 2024-05-05
fix: terminate early on first counterexample with --early-exit

### LINKED ISSUE #243: --early-exit doesn't work as expected
**Describe the bug**

`--early-exit` is supposed to print the first counterexample found and stop halmos, but we can actually still have multiple counterexamples reported

**To Reproduce**

```
pragma solidity ^0.8.15;
import "forge-std/Test.sol";

contract Test60 is Test {
    function test_manyCexes(uint256 x) external {
        uint256 i = 0;
        if (x & 1 == 0) { ++i; }
        if (x & 2 == 0) { ++i; }
        if (x & 4 == 0) { ++i; }
        if (x & 8 == 0) { ++i; }

        assert(x < 64);
    }
}
```

```
halmos --function test_manyCexes --early-exit

Running 1 tests for test/60_manyCexes.t.sol:Test60
Counterexample: 
    p_x_uint256 = 0x8000000000000000000000000000000000000000000000000000000000000000 (57896044618658097711785492504343953926634992332820282019728792003956564819968)
Counterexample: 
    p_x_uint256 = 0x8000000000000000000000000000000000000000000000000000000000000008 (57896044618658097711785492504343953926634992332820282019728792003956564819976)
Counterexample: 
    p_x_uint256 = 0x8000000000000000000000000000000000000000000000000000000000000004 (57896044618658097711785492504343953926634992332820282019728792003956564819972)
Counterexample: 
    p_x_uint256 = 0x800000000000000000000000000000000000000000000000000000000000000c (57896044618658097711785492504343953926634992332820282019728792003956564819980)
...
```

saurabhchalke:
## Description

This pull request fixes the issue with the `--early-exit` flag not terminating Halmos immediately after finding the first counterexample, as reported in #243.

## Changes

The following changes have been made to address the issue:

1. Introduced an `early_exit` flag variable to keep track of whether an early exit is requested.

2. In the `future_callback` function, when a counterexample is found and the `--early-exit` flag is set, the `early_exit` flag is set to `True`.

3. The `future_callback` function now checks the `early_exit` flag at the beginning, and if it is `True`, it immediately returns without processing the result. This prevents any further counterexamples from being processed after an early exit is requested.

4. The thread pool is shut down after all the submitted futures have completed or been canceled, ensuring a clean termination.

## Issue

Closes #243

## Testing

The updated code has been tested with the example contract provided in the issue description. Here's the updated output:

```bash
(.venv) 86fb4ce7caf3% python3 -m halmos --function test_manyCexes --early-exit
[⠊] Compiling...
[⠢] Compiling 28 files with 0.8.25
[⠔] Solc 0.8.25 finished in 2.15s
Compiler run successful!
Running 1 tests for test/Test60.t.sol:Test60
Counterexample:
p_x_uint256 = 0x8000000000000000000000000000000000000000000000000000000000000008 (57896044618658097711785492504343953926634992332820282019728792003956564819976)
[FAIL] test_manyCexes(uint256) (paths: 32, time: 0.08s, bounds: [])
Symbolic test result: 0 passed; 1 failed; time: 0.09s
```

As demonstrated, with the `--early-exit` flag set, Halmos now reports only the first counterexample found and terminates early.

Please review the changes and provide any feedback or suggestions. @karmacoma-eth @daejunpark

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -65,6 +65,8 @@ VERBOSITY_TRACE_SETUP = 3
 VERBOSITY_TRACE_PATHS = 4
 VERBOSITY_TRACE_CONSTRUCTOR = 5
 
+solver_contexts = []
+
 
 @dataclass(frozen=True)
 class FunctionInfo:
@@ -657,7 +659,14 @@ def run(
     counterexamples = []
     traces = {}
 
+    early_exit = False
+
     def future_callback(future_model):
+        nonlocal early_exit
+
+        if early_exit:
+            return
+
         m = future_model.result()
         models.append(m)
 
@@ -665,17 +674,21 @@ def run(
         if result == unsat:
             return
 
+        counterexample_found = False
+
         # model could be an empty dict here
         if model is not None:
             if is_valid:
                 print(red(f"Counterexample: {render_model(model)}"))
                 counterexamples.append(model)
+                counterexample_found = True
             else:
                 warn(
                     COUNTEREXAMPLE_INVALID,
                     f"Counterexample (potentially invalid): {render_model(model)}",
                 )
                 counterexamples.append(model)
+                counterexample_found = True
         else:
             warn(COUNTEREXAMPLE_UNKNOWN, f"Counterexample: {result}")
 
@@ -690,6 +703,26 @@ def run(
                 else "Trace:"
             )
             print(traces[index], end="")
+
+        if args.early_exit and counterexample_found:
+            early_exit = True
+
+            # Interrupt all the solver contexts
+            for ctx in solver_contexts:
+                try:
+                    ctx.interrupt()
+                except Exception as e:
+                    if args.debug:
+                        traceback.print_exc()
+
+            # Cancel the remaining futures
+            for future in future_models:
+                if not future.done():
+                    future.cancel()
+
+            # Clear the global lists to prevent memory leaks
+            solver_contexts.clear()
+            future_models.clear()
 
     for idx, ex in enumerate(exs):
         result_exs.append(ex)
@@ -739,16 +772,7 @@ def run(
             f"# of potential paths involving assertion violations: {len(future_models)} / {len(result_exs)}  (--solver-threads {args.solver_threads})"
         )
 
-    if args.early_exit:
-        while not (
-            len(counterexamples) > 0 or all([fm.done() for fm in future_models])
-        ):
-            time.sleep(1)
-
-        thread_pool.shutdown(wait=False, cancel_futures=True)
-
-    else:
-        thread_pool.shutdown(wait=True)
+    thread_pool.shutdown(wait=True)
 
     counter = Counter(str(m.result) for m in models)
     if counter["sat"] > 0:
@@ -1054,6 +1078,7 @@ def solve(
         solver.from_string(query)
         result = solver.check()
         model = copy_model(solver.model()) if result == sat else None
+        solver_contexts.append(solver.ctx)
         return result, model
 
 


karmacoma-eth:
thanks for taking a stab at this! I'm taking a look

karmacoma-eth:
ok so I think the problem with this is that while it does improve the output, it does not actually exit early. 

To demonstrate, we can make the assertion in manyCexes harder:

```solidity
pragma solidity ^0.8.15;

import "forge-std/Test.sol";

contract Test60 is Test, SymTest {
    function test_manyCexes(uint256 x) external {
        uint256 i = 0;
        if (x & 1 == 0) ++i;
        if (x & 2 == 0) ++i;
        if (x & 4 == 0) ++i;
        if (x & 8 == 0) ++i;

        uint256 y = svm.createUint256("y");
        uint256 z = svm.createUint256("z");
        assertNotEq(y * z, 4096);  // can make this harder by changing the value here
    }
}
```

What happens is this:
1. we spawn many solver threads (the hard work is in `gen_model_from_sexpr`)
2. one thread finds an answer, its `future_callback` runs and sets `counterexample_found = True` and `early_exit = True`
3. all the other threads keep running (⚠️ this is the issue)
4. when they also find the counterexample, they don't print it because `early_exit` has been set to True

If you could help us figure out how to safely interrupt the other running threads as soon as a counterexample is found that would be amazing

karmacoma-eth:
apparently we can't interrupt tasks that are running in `ThreadPoolExecutor`, but we _can_ interrupt a solver context (`solver.ctx.interrupt()`), we could try that out

karmacoma-eth (changes requested):
see other comments, let's figure out how to solve the underlying problem before merging 🙏

saurabhchalke:
Thank you @karmacoma-eth for the guidance and valuable feedback! Your suggestions have been really helpful in identifying the underlying issue and finding the right approach to solve it.

I have updated the code based on your recommendations. Here's a summary of the changes:

- Introduced a global list `solver_contexts` to store the solver contexts created in the `solve` function.
- Modified the `solve` function to append the solver context to the global list before returning.
- In the `future_callback` function, when `args.early_exit` is true and a counterexample is found, all the solver contexts in the global list are interrupted using `ctx.interrupt()`.
- Added exception handling to the interruption process to avoid segmentation faults.
- After interrupting the solvers, the global list is cleared to prevent memory leaks.

With these changes, the `--early-exit` flag should now work as expected, interrupting all the running solver contexts and allowing Halmos to terminate early when a counterexample is found.

I have a question regarding the use of `ThreadPoolExecutor` here. Have you considered using `ProcessPoolExecutor` instead? Since the solver computations are CPU-bound tasks, using separate processes instead of threads might lead to better performance and resource utilization. It would be interesting to explore this option and see if it brings any improvements.

Please let me know if you have any further suggestions or if there's anything else I can improve in this pull request.

Thank you for your time and support!

karmacoma-eth:
unfortunately this still doesn't seem to work as expected. I see the calls to `ctx.interrupt()`, but the other threads keep running anyway.

Here is another example that really shows what we need, it generates increasingly difficult problems (for z3 at least) and we want to see the runtime improved, it should really exit as soon as the first counterexample is found:

```sol
pragma solidity ^0.8.15;

import "forge-std/Test.sol";

contract Test60 is Test, SymTest {
    /// @custom:halmos --loop 12
    function test_manyCexes() external {
        for(uint256 i = 1; i <= 12; ++i) {
            uint256 y = svm.createUint256("y");
            uint256 z = svm.createUint256("z");
            assertNotEq(y * z, 1 << i);
        }
    }
}
```

karmacoma-eth:
@saurabhchalke thank you for the attempt, I'll close this PR since unfortunately ctx.interrupt() doesn't cut it. It looks like the solution will be to move away from a ThreadPoolExecutor and use external processes instead

saurabhchalke:
@karmacoma-eth Thank you for your feedback and for highlighting the limitations with ctx.interrupt(). I understand the need to move away from ThreadPoolExecutor and appreciate the guidance.

PR #292, opened 2024-05-22
perf: improve branching solving

daejunpark:
wip


daejunpark:
close this, as branching solving is needed for current loop detection logic.

PR #293, opened 2024-05-22
Optimize execution loop 

karmacoma-eth:
Completely driven by flamegraphs, I kept squeezing until the fetch-decode-execute loop almost disappeared from flamegraphs (now dominated by other things like calls to `resolve_address_alias`):

Tested on morpho blue, 24.8% faster:

    before: check_fee(bytes4,address) (paths: 38, time: 21.82s (paths: 21.82s, models: 0.00s), bounds: [])
    after:  check_fee(bytes4,address) (paths: 39, time: 16.41s (paths: 16.41s, models: 0.00s), bounds: [])

Avoids duplicating work:

- load pgm[pc] to know the current instruction (doing a chunk lookup + indexing in the code bytevec)
- load pgm[pc+1:pc+1+operand_size] for PUSHn instructions (every time it is executed, which can be expensive)
- load pgm[pc] again to know the next_pc

Instead we do this eagerly once and then remember the result. 

Plus some extra goodies:
- support for MCOPY 
- explicit Message.call_scheme, which caused us to render deploy output in traces at high verbosity when the intent was to just display the number of bytes


karmacoma-eth:
Tested on Farcaster, 11% faster:

```
# before
[PASS] check_Invariants_PostMigration(bytes4,address) (paths: 641, time: 195.43s (paths: 195.43s, models: 0.00s), bounds: [])

# after
[PASS] check_Invariants_PostMigration(bytes4,address) (paths: 637, time: 173.95s (paths: 173.95s, models: 0.00s), bounds: [])
```

src/halmos/sevm.py
```diff
@@ -2097,17 +2094,11 @@ def finalize(ex: Exec):
...
-
                 opcode = insn.opcode
-                ex.cnts["opcode"][opcode] += 1
 
-                if (
-                    "max_depth" in self.options
-                    and sum(ex.cnts["opcode"].values()) > self.options["max_depth"]
-                ):
+                max_depth = self.options.get("max_depth", 0)
+                if max_depth and step_id > max_depth:
```
daejunpark, line 2101: unlike cnts, step_id is intra-procedural, not accumulating the steps of external callees.

to make this optimization correct, we need to have step_id to be accumulated across external calls.

src/halmos/sevm.py
```diff
@@ -2407,6 +2398,21 @@ def finalize(ex: Exec):
...
+                    if size > 0:
+                        src_end_loc = src_offset + size
+                        dst_end_loc = dest_offset + size
+
+                        if max(src_end_loc, dst_end_loc) > MAX_MEMORY_SIZE:
+                            raise HalmosException("MCOPY > MAX_MEMORY_SIZE")
+
+                        data = ex.st.memory.slice(src_offset, src_end_loc)
+                        ex.st.memory.set_slice(dest_offset, dst_end_loc, data)
+
```
daejunpark, line 2415: just to double-check, does this also work as expected when the src range overlaps with the dest range?

src/halmos/utils.py
```diff
@@ -314,7 +317,7 @@ def hexify(x):
     if isinstance(x, str):
         return re.sub(r"\b(\d+)\b", lambda match: hex(int(match.group(1))), x)
     elif isinstance(x, int):
-        return hex(x)
+        return f"0x{x:02x}"
```
daejunpark, line 320: just curious, is this only about formatting, or also about (micro-)performance?

src/halmos/utils.py
```diff
@@ -257,18 +257,21 @@ def unbox_int(x: Any) -> Any:
...
 
-    if is_bv(x):
-        x = simplify(x)
-
     if is_bv_value(x):
         return x.as_long()
 
+    if is_bv(x):
+        x = simplify(x)
+
```
daejunpark, line 274: nice reordering! impressed that profiling is so much helpful to reveal this.

.github/workflows/test-external.yml
```diff
@@ -22,11 +22,11 @@ jobs:
           - repo: "a16z/cicada"
             dir: "cicada"
             cmd: "halmos --contract LibUint1024Test --function testProve --loop 256"
-            branch: ""
+            branch: "pin-solc"
           - repo: "a16z/cicada"
             dir: "cicada"
             cmd: "halmos --contract LibPrimeTest --function testProve --loop 256"
-            branch: ""
+            branch: "pin-solc"
```
daejunpark, line 29: 😢

.github/workflows/test-external.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
daejunpark, line 12: any reason to switch to ubuntu?

oh, when i checked last time, macos-latest provided more cpus/memory, but it's no longer the case [now](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories):
- ubuntu-latest: 4 cores / 16 gb
- macos-latest: 3 (M1) cores / 7 gb

daejunpark: if ubuntu-latest is better, should we switch the test-long task to ubuntu as well?

.github/workflows/test-external.yml
```diff
@@ -67,5 +73,5 @@ jobs:
         run: python -m pip install -e ./halmos
 
       - name: Test external repo
-        run: ${{ matrix.cmd }} -v -st --solver-timeout-assertion 0 --solver-threads 2
+        run: ${{ matrix.cmd }} -v -st --solver-timeout-assertion 0 --solver-threads 6 --solver-command=yices-smt2
```
daejunpark, line 76: i'm concerned too many parallel solver processes could actually degrade performance. we have only 4 cores in ubuntu runner. wdyt?

src/halmos/sevm.py
```diff
@@ -861,19 +862,15 @@ def returndatasize(self) -> int:
...
 
     def jumpi_id(self) -> str:
-        return f"{self.pc}:" + ",".join(
-            map(lambda x: str(x) if self.is_jumpdest(x) else "", self.st.stack)
-        )
+        # no need to scan the entire stack, limit ourselves to the top N elements
+        # limiting ourselves to the top 16 leads to false positives
+        top_stack = self.st.stack[:32]
+        jumpdests_str = map(lambda x: str(x) if self.is_jumpdest(x) else "", top_stack)
+        return f"{self.pc}:{','.join(jumpdests_str)}"
```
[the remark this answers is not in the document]
karmacoma-eth: > could you elaborate why we don't need to scan the entire stack?

this is a costly operation, and since the loop detection is a heuristic I am looking for a good trade-off. I can try to think of a more reliable alternative

src/halmos/sevm.py
karmacoma-eth: not sure what you mean, I see step_id incrementing monotonically along a single path. This may slightly change the meaning of `max_depth`, but the new behavior sounds reasonable to me

image

daejunpark: oops, sorry my bad. a separate sevm is no longer spawn for external calls, since the new dfs path exploration. so yes, step_id increases across external calls. great catch, thanks!

src/halmos/sevm.py
karmacoma-eth: knowing how this works I want to say clearly yes, but I actually added tests for it in b255289, can't hurt

src/halmos/utils.py
karmacoma-eth: just formatting

.github/workflows/test-external.yml
karmacoma-eth: yes might be worth it, for me the driver was faster yices install but more memory and cores is nice too. Although I think eventually we probably want to migrate everything to the Docker workflow

.github/workflows/test-external.yml
karmacoma-eth: you're right, back to 4

.github/workflows/test-external.yml
karmacoma-eth: related issue here: https://github.com/ethereum/solidity/issues/14358

src/halmos/sevm.py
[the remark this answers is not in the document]
daejunpark: yes i agree it is costly. and yes it is heuristic anyway, but new heuristic requires another validation process / period, which deserves its own pr and separate discussion.

a possible optimization without changing heuristic is to maintain the jumpi id inside the stack, but this could be nontrivial to implement.

i'd suggest not spending too much time on this loop detection heuristic, as we might want to develop a different loop detection, especially for the new branching condition solving.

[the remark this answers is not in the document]
karmacoma-eth: good point, I'll revert this particular change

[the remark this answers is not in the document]
karmacoma-eth: back to whole stack scanning in eac7662

daejunpark (approved):

PR #295, opened 2024-05-22
feat: use docker image to load solvers and setup the testing environment

EmperorOrokuSaki:
This pull request uses the [solvers](https://github.com/EmperorOrokuSaki/solvers/pkgs/container/solvers) package to load a docker container that has various solvers and dependencies of Halmos setup.


karmacoma-eth:
thanks @EmperorOrokuSaki, the functionality looks good and I can confirm it works in our CI here:
https://github.com/a16z/halmos/actions/runs/9198938639

since it works, I believe we can merge it as-is but if you're up for it it would be awesome to have a follow-up PR with:

- if possible, two split images: 
    - a `solvers` image that is just ubuntu + yices, z3, bitwuzla, cvc5, etc. (as many solvers as feasible). No python, foundry or halmos, just the smallest image that contains solvers.
    - a `halmos` image that uses `solvers` as a base image and has Python, foundry and the latest halmos

- the Dockerfiles in-tree. Since we're going to rely on this in CI, we need to be able to easily update the image with updated dependencies

karmacoma-eth (approved):

PR #299, opened 2024-06-01
example halmos.toml

karmacoma-eth:
can generate a sample config file using:

    python -m halmos.config ARGS > halmos.toml

uses the help text, resolved values, metavars and type info 🤠

### DIFF
--- a/src/halmos/config.py
+++ b/src/halmos/config.py
@@ -88,7 +88,7 @@ class Config:
 
     config: str = arg(
         help="path to the configuration file",
-        global_default=None,
+        global_default=os.path.join(os.getcwd(), "halmos.toml"),
         metavar="FILE",
     )
 
@@ -98,7 +98,7 @@ class Config:
 
     contract: str = arg(
         help="run tests in the given contract. Shortcut for `--match-contract '^{NAME}$'`.",
-        global_default=None,
+        global_default="",
         metavar="CONTRACT_NAME",
     )
 
@@ -129,13 +129,13 @@ class Config:
     )
 
     width: int = arg(
-        help="set the max number of paths, 0 for unlimited",
+        help="set the max number of paths; 0 means unlimited",
         global_default=0,
         metavar="MAX_WIDTH",
     )
 
     depth: int = arg(
-        help="set the maximum length in steps of a single path, 0 for unlimited",
+        help="set the maximum length in steps of a single path; 0 means unlimited",
         global_default=0,
         metavar="MAX_DEPTH",
     )
@@ -294,7 +294,7 @@ class Config:
     ### Build options
 
     forge_build_out: str = arg(
-        help="forge build artifacts directory name (default: 'out/')",
+        help="forge build artifacts directory name",
         metavar="DIRECTORY_NAME",
         global_default="out",
         group=build,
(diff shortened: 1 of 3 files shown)


tests/regression/halmos.toml
```diff
@@ -0,0 +1,148 @@
...
+solver-max-memory = 0
+
+# use the given command when invoking the solver
+# solver-command = COMMAND
+
+# run assertion solvers in parallel
+solver-parallel = false
+
+# set the number of threads for parallel solvers
+solver-threads = 20
```
daejunpark, line 132:
```suggestion
# solver-threads = N
```

daejunpark: the value for --solver-threads doesn't need to be set here, as it can be set at runtime based on the runner's cpu info.

can we have the auto-generation to set solver-threads only when it is explicitly given by command line?

src/halmos/config.py
```diff
@@ -611,8 +611,67 @@ def toml_parser():
...
+    # devs can have a little easter egg
+    lines = [
+        "#     ___       ___       ___       ___       ___       ___",
+        "#    /\\__\\     /\\  \\     /\\__\\     /\\__\\     /\\  \\     /\\  \\",
+        "#   /:/__/_   /::\\  \\   /:/  /    /::L_L_   /::\\  \\   /::\\  \\",
+        "#  /::\\/\\__\\ /::\\:\\__\\ /:/__/    /:/L:\\__\\ /:/\\:\\__\\ /\\:\\:\\__\\",
+        "#  \\/\\::/  / \\/\\::/  / \\:\\  \\    \\/_/:/  / \\:\\/:/  / \\:\\:\\/__/",
+        "#    /:/  /    /:/  /   \\:\\__\\     /:/  /   \\::/  /   \\::/  /",
+        "#    \\/__/     \\/__/     \\/__/     \\/__/     \\/__/     \\/__/",
+    ]
```
daejunpark, line 637: 😍

src/halmos/config.py
```diff
@@ -88,7 +88,7 @@ class Config:
 
     config: str = arg(
         help="path to the configuration file",
-        global_default=None,
+        global_default=os.path.join(os.getcwd(), "halmos.toml"),
```
daejunpark, line 91: if i understand correctly, this default value is only used when generating the help message, as `resolve_config_files()` doesn't use this default value, right?

in that case, here can we use a meta variable, say `ROOT`, instead of `os.getcwd()`, to highlight that the config file will be searched under the root, which is not necessarily the current directory?

daejunpark (approved):

tests/regression/halmos.toml
karmacoma-eth: yes you're right

src/halmos/config.py
karmacoma-eth: it now says:

```
options:
  -h, --help            show this help message and exit
  --root ROOT           project root directory (default: current working directory)
  --config FILE         path to the config file (default: ROOT/halmos.toml)
```

karmacoma-eth: > this default value is only used when generating the help message

yeah that's right

PR #300, opened 2024-06-01
🔨 Add Missing Error Code in `parse_build_out` Warning

pcaversaccio:
Patching a refactor regression from https://github.com/a16z/halmos/pull/298.


karmacoma-eth (approved):
thanks for fixing, and sorry I didn't catch it in testing 🙈

PR #307, opened 2024-06-07
run workflow using `solvers` as a base image

karmacoma-eth:
Example run: 
https://github.com/a16z/halmos/actions/runs/9409762438

This works at least as a proof of concept. We should be able to optimize the 1min "build image step" a little bit more:
- since we want to keep the solvers package very minimal, we need to add the other dependencies we need in CI for halmos (git, python, foundry, etc)
- we need to checkout the current halmos from github and install it

It would be faster to use either an already published `halmos:latest` image (but wouldn't help for running tests on a branch). 

Alternatively we could also publish an intermediary `halmos-base` image that already includes git, python, foundry, uv, etc. and we would only need to install the halmos checkout.

### DIFF
--- a/.github/workflows/test-long.yml
+++ b/.github/workflows/test-long.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: Install Git
-        run: sudo apt-get install -y git
+        run: apt-get install -y git
 
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


karmacoma-eth:
(note: this PR does not yet enable publishing the `halmos` package, will do separately)

daejunpark (approved):

daejunpark:
i ran again the test-ffi but it kept failing. do you have any idea?

karmacoma-eth:
> i ran again the test-ffi but it kept failing. do you have any idea?

weird, 404 trying to download foundry. Maybe now is a good time to switch test-ffi to the docker image too, since it doesn't rely on downloading the github release :)

PR #308, opened 2024-06-07
add intermediary halmos-builder image and dockerize more workflows

karmacoma-eth:
halmos-builder:
- based on `solvers`
- adds dev dependencies like python, git, wget, foundry, uv, etc.

halmos:
- based on `halmos-builder`, just does a `uv pip install` of a halmos checkout

For test-long and test-ffi, we:
- check out the latest halmos commit
- build a new `halmos` image based on that
- runs pytest inside the resulting `halmos` container


karmacoma-eth:
note: we may eventually want to have a `halmos-dev` package that contains pytest and everything necessary to run halmos tests in our own workflows (~600MB image), and a slim `halmos` package without the tests folder, pytest, etc (can probably get it down to <300MB)

.github/workflows/test-external.yml
```diff
@@ -22,43 +24,59 @@ jobs:
...
           submodules: false
 
+      - name: Build image
+        run: docker build -t halmos-image . --file packages/halmos/Dockerfile
+
+      - name: Install Vyper
+        if: ${{ matrix.project.dir == 'snekmate' }}
+        run: |
+          docker run --name halmos-vyper --entrypoint uv halmos-image pip install vyper
+          docker commit --change 'ENTRYPOINT ["halmos"]' halmos-vyper halmos-image
```
karmacoma-eth, line 78: shows that we can layer vyper on top of the existing halmos image fairly easily in github workflows

.github/workflows/test-external.yml
```diff
@@ -22,43 +24,59 @@ jobs:
         project:
           - repo: "morpho-org/morpho-data-structures"
             dir: "morpho-data-structures"
-            cmd: "halmos --function testProve --loop 4 --symbolic-storage"
+            cmd: "--function testProve --loop 4 --symbolic-storage"
```
daejunpark, line 27: is the yices solver not used here due to the weird failure you mentioned?

.github/workflows/test-external.yml
```diff
@@ -22,43 +24,59 @@ jobs:
...
           # we won't be needing tests/lib for this workflow
           submodules: false
 
+      - name: Build image
+        run: docker build -t halmos-image . --file packages/halmos/Dockerfile
+
+      - name: Install Vyper
+        if: ${{ matrix.project.dir == 'snekmate' }}
+        run: |
+          docker run --name halmos-vyper --entrypoint uv halmos-image pip install vyper
```
daejunpark, line 77: would this work as well?  having `halmos-image` placed between `uv` and `pip install vyper` is a bit confusing.  wdyt?
```suggestion
          docker run --name halmos-vyper halmos-image --entrypoint uv pip install vyper
```

.github/workflows/test-external.yml
```diff
@@ -67,35 +85,11 @@ jobs:
...
-      - name: Install Yices 2 SMT solver
-        run: |
-          wget https://github.com/SRI-CSL/yices2/releases/download/Yices-2.6.4/yices-2.6.4-x86_64-pc-linux-gnu.tar.gz
-          tar -xzvf yices-2.6.4-x86_64-pc-linux-gnu.tar.gz
-          sudo mv yices-2.6.4/bin/* /usr/local/bin/
-          sudo mv yices-2.6.4/lib/* /usr/local/lib/
-          sudo mv yices-2.6.4/include/* /usr/local/include/
-          rm -rf yices-2.6.4
+      - name: Print halmos version
+        run: docker run -v .:/workspace halmos-image --version
```
daejunpark, line 89: the halmos version printed here is quite different from what we get locally.
```
Run docker run -v .:/workspace halmos-image --version
halmos 0.1.dev1+g81b7ae5
```
https://github.com/a16z/halmos/actions/runs/10014074545/job/27683005509

can we improve the dockerfile to contain the correct halmos version?

.github/workflows/test-external.yml
karmacoma-eth: yes :(

.github/workflows/test-external.yml
karmacoma-eth: no, it's a weird docker command line quirk. I agree it's confusing but I couldn't find a workaround, it really insists on this format

`Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]`

karmacoma-eth: if you invoke `docker run --name halmos-vyper halmos-image --entrypoint uv pip install vyper`, it will invoke the current entrypoint of `halmos-image` with the rest of the arguments, so it will run `halmos --entrypoint uv pip install vyper`

.github/workflows/test-external.yml
karmacoma-eth: the only difference I can see is that we install halmos via `uv pip install` in the Dockerfile instead of `pip install`

.github/workflows/test-external.yml
daejunpark: oh i see, TIL!

.github/workflows/test-external.yml
karmacoma-eth: oh and actually the other big difference is the sparse checkout we do in github actions, which explains the `0.1.dev1` part. I think it's fine, the git hash has the information we need really

daejunpark (approved):

PR #309, opened 2024-06-11
[Feature/#244] Parse symbols from build to print readable traces

### LINKED ISSUE #244: Parse symbols from build to print readable traces
**Is your feature request related to a problem? Please describe.**

Traces currently display only EVM-level raw hex values (for logs, calldata, deployed contracts, etc). 

**Describe the solution you'd like**

It would be nice to see the name of events, function calls and deployed contracts instead when these are available in the build output.

pillip:
## Change Type

- [ ] Bug fix (Not Breaking-change. Solve issue.)
- [x] New feature (Not Breaking-change. Add new feature.)
- [ ] Breaking change (Bug fix or New featrure that cause Breaking change.)
- [ ] Code Convention (Edit Code Convention.)
- [ ] Refactoring (Refactor Code without feature change.)

## Description
This PR introduces several significant updates and new functionalities that parse and covert raw hex code to human readable string. The changes primarily focus on enhancing the Mapper class with methods for AST parsing and adding the DeployAddressMapper class for managing deployed contract address mappings.

## Related Issues
This PR resolves #244 

## Changes
1. Mapper class
    * Implemented using the Singleton pattern to manage contract and node mapping information from AST
    * Use AstNode and ContractMappingInfo dataclasses
    * Methods
        * parse_ast: Recursively parses AST nodes.
        * add_contract_mapping_info: Adds a new contract mapping
        * get_contract_mapping_info_by_name: Retrieves contract mapping by contract name
        * get_contract_mapping_info_by_bytecode: Retrieves contract mapping by bytecode suffix
        * append_node: Appends a new AST node to contract mapping info
        * find_nodes_by_address: Converts hex addresses to human-readable strings including contract name, event name, function name, and variable name

2. DeployAddressMapper class
    * Implemented using the Singleton pattern
    * Added to manage deployed contract address mappings with predefined mappings such as HEVM and SVM.

3. Modify functions when used for printing traces
    * render_trace: Added logic to look up deployed contract names.
    * rendered_calldata, hexify, find_nodes_by_address: Updated to support contract name parameters for searching nodes in specific contracts.

## Tests
```
> cd halmos/examples/simple
> halmos -vvvv
```

## Screenshot
image

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -9,7 +9,6 @@ import sys
 import time
 import traceback
 import uuid
-
 from collections import Counter
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from dataclasses import asdict, dataclass
@@ -18,13 +17,9 @@ from importlib import metadata
 
 from .bytevec import ByteVec
 from .calldata import Calldata
-from .config import (
-    arg_parser,
-    default_config,
-    resolve_config_files,
-    toml_parser,
-    Config as HalmosConfig,
-)
+from .config import Config as HalmosConfig
+from .config import arg_parser, default_config, resolve_config_files, toml_parser
+from .mapper import Mapper
 from .sevm import *
 from .utils import (
     NamedTimer,
@@ -1279,6 +1274,29 @@ def parse_build_out(args: HalmosConfig) -> Dict:
                         sol_dirname,
                     )
                 contract_map[contract_name] = (json_out, contract_type, natspec)
+
+                try:
+                    bytecode = contract_map[contract_name][0]["bytecode"]["object"]
+                    contract_mapping_info = Mapper().get_contract_mapping_info_by_name(
+                        contract_name
+                    )
+
+                    if contract_mapping_info is None:
+                        Mapper().add_contract_mapping_info(
+                            contract_name=contract_name,
+                            bytecode=bytecode,
+                            nodes=[],
+                        )
+                    else:
+                        contract_mapping_info.bytecode = bytecode
+
+                    contract_mapping_info = Mapper().get_contract_mapping_info_by_name(
+                        contract_name
+                    )
+                    Mapper().parse_ast(contract_map[contract_name][0]["ast"])
+
+                except Exception:
+                    pass
             except Exception as err:
                 warn_code(
                     PARSING_ERROR,
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -19,7 +19,7 @@ from .bytevec import ByteVec
 from .calldata import Calldata
 from .config import Config as HalmosConfig
 from .config import arg_parser, default_config, resolve_config_files, toml_parser
-from .mapper import Mapper
+from .mapper import DeployAddressMapper, Mapper
 from .sevm import *
 from .utils import (
     NamedTimer,
@@ -276,18 +276,15 @@ def rendered_trace(context: CallContext) -> str:
         return output.getvalue()
 
 
-def rendered_calldata(calldata: ByteVec) -> str:
-    return hexify(calldata.unwrap()) if calldata else "0x"
+def rendered_calldata(calldata: ByteVec, contract_name: str = None) -> str:
+    return hexify(calldata.unwrap(), contract_name) if calldata else "0x"
 
 
 def render_trace(context: CallContext, file=sys.stdout) -> None:
-    # TODO: label for known addresses
-    # TODO: decode calldata
-    # TODO: decode logs
-
     message = context.message
     addr = unbox_int(message.target)
     addr_str = str(addr) if is_bv(addr) else hex(addr)
+    addr_str = DeployAddressMapper().get_deployed_contract(addr_str)
 
     value = unbox_int(message.value)
     value_str = f" (value: {value})" if is_bv(value) or value > 0 else ""
@@ -298,13 +295,29 @@ def render_trace(context: CallContext, file=sys.stdout) -> None:
     if message.is_create():
         # TODO: select verbosity level to render full initcode
         # initcode_str = rendered_initcode(context)
+
+        try:
+            if context.output.error is None:
+                target = hex(int(str(message.target)))
+                bytecode = context.output.data.unwrap().hex()
+                contract_name = (
+                    Mapper()
+                    .get_contarct_mapping_info_by_bytecode(bytecode)
+                    .contract_name
+                )
+
+                DeployAddressMapper().add_deployed_contract(target, contract_name)
+                addr_str = contract_name
+        except:
+            pass
+
         initcode_str = f"<{byte_length(message.data)} bytes of initcode>"
         print(
             f"{indent}{call_scheme_str}{addr_str}::{initcode_str}{value_str}", file=file
         )
 
     else:
-        calldata = rendered_calldata(message.data)
+        calldata = rendered_calldata(message.data, addr_str)
         call_str = f"{addr_str}::{calldata}"
         static_str = yellow(" [static]") if message.is_static else ""
         print(f"{indent}{call_scheme_str}{call_str}{static_str}{value_str}", file=file)
@@ -1519,6 +1532,9 @@ def _main(_args=None) -> MainResult:
 
         contract_path = f"{contract_json['ast']['absolutePath']}:{contract_name}"
         print(f"\nRunning {num_found} tests for {contract_path}")
+
+        # Set 0xaaaa0001 in DeployAddressMapper
+        DeployAddressMapper().add_deployed_contract("0xaaaa0001", contract_name)
 
         # support for `/// @custom:halmos` annotations
         contract_args = with_natspec(args, contract_name, natspec)
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -303,7 +303,7 @@ def render_trace(context: CallContext, file=sys.stdout) -> None:
                 bytecode = context.output.data.unwrap().hex()
                 contract_name = (
                     Mapper()
-                    .get_contarct_mapping_info_by_bytecode(bytecode)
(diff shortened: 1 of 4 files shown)


src/halmos/__main__.py
```diff
@@ -281,18 +276,15 @@ def rendered_trace(context: CallContext) -> str:
...
 
 def render_trace(context: CallContext, file=sys.stdout) -> None:
-    # TODO: label for known addresses
-    # TODO: decode calldata
-    # TODO: decode logs
-
     message = context.message
     addr = unbox_int(message.target)
     addr_str = str(addr) if is_bv(addr) else hex(addr)
+    addr_str = DeployAddressMapper().get_deployed_contract(addr_str)
```
pillip, line 287: For converting from address to contract name

src/halmos/mapper.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@dataclass
+class ContractMappingInfo:
+    contract_name: str
+    bytecode: str
+    nodes: List[AstNode]
+
+
+class SingletonMeta(type):
```
pillip, line 21: a metaclass for implementing the Singleton pattern

src/halmos/utils.py
```diff
@@ -313,23 +316,25 @@ def decode_hex(hexstring: str) -> Optional[bytes]:
         return None
 
 
-def hexify(x):
+def hexify(x, contract_name: str = None):
```
pillip, line 319: By setting the default value of contract_name to None, existing usage of this function can use without modification

src/halmos/mapper.py
```diff
@@ -0,0 +1,173 @@
...
+        self._contracts[contract_name] = ContractMappingInfo(
+            contract_name, bytecode, nodes
+        )
+
+    def get_contract_mapping_info_by_name(
+        self, contract_name: str
+    ) -> Optional[ContractMappingInfo]:
+        return self._contracts.get(contract_name, None)
+
+    def get_contarct_mapping_info_by_bytecode(
```
karmacoma-eth, line 58:
```suggestion
    def get_contract_mapping_info_by_bytecode(
```

src/halmos/mapper.py
```diff
@@ -0,0 +1,173 @@
...
+        self, bytecode: str
+    ) -> Optional[ContractMappingInfo]:
+        for contract_mapping_info in self._contracts.values():
+            if contract_mapping_info.bytecode.endswith(bytecode):
```
karmacoma-eth, line 62: I don't think this works for contracts with immutable variables. It's fine for now, can leave it as a TODO

karmacoma-eth:
this looks great to me, thanks @pillip!

src/halmos/__main__.py
karmacoma-eth:
```suggestion
    # check if we have a contract name for this address in our deployment mapper
    addr_str = DeployAddressMapper().get_deployed_contract(addr_str)
```

karmacoma-eth: Quick test to see how foundry deals with the same contract deployed at multiple addresses:

This:

```solidity
contract CounterTest is Test {
    function setUp() public {
        Counter counter = new Counter();
        counter.setNumber(0);

        Counter counter1 = new Counter();
        counter1.setNumber(1);

        Counter counter2 = new Counter();
        counter2.setNumber(2);
    }
}
```

Is rendered like this:

```
[PASS] testIncrement() (gas: 28334)
Traces:
  [315303] CounterTest::setUp()
    ├─ [49499] → new Counter@0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f
    │   └─ ← [Return] 247 bytes of code
    ├─ [2390] Counter::setNumber(0)
    │   └─ ← [Stop] 
    ├─ [49499] → new Counter@0x2e234DAe75C793f67A35089C9d99245E1C58470b
    │   └─ ← [Return] 247 bytes of code
    ├─ [22290] Counter::setNumber(1)
    │   └─ ← [Stop] 
    ├─ [49499] → new Counter@0xF62849F9A0B5Bf2913b396098F7c7019b51A820a
    │   └─ ← [Return] 247 bytes of code
    ├─ [22290] Counter::setNumber(2)
    │   └─ ← [Stop] 
    └─ ← [Stop] 
```

I don't love the ambiguity about which Counter instance is called, but we're on par with foundry so fine to keep it as is for now

src/halmos/mapper.py
karmacoma-eth: quick confirm, adding this to BadElections.sol:

```solidity
    address immutable public owner;

    constructor() {
        owner = msg.sender;
    }
```

results:

```
# before
STATICCALL BadElections::Concat(votesFor, p_proposalId_uint256()) [static]

# after
STATICCALL 0xaaaa0002::Concat(BadElections.votesFor, p_proposalId_uint256()) [static]
```

src/halmos/__main__.py
pillip: @karmacoma-eth 
Thank you for your feedback!
I hadn't considered this example. Now I understand the concern about the ambiguity regarding which Counter instance is called.

src/halmos/mapper.py
pillip: @karmacoma-eth 
Thank you for feedback!
I added TODO comment at commit 0611dd4

src/halmos/mapper.py
```diff
@@ -0,0 +1,181 @@
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Type
+
+
+@dataclass
+class AstNode:
+    node_type: str
+    id: int
+    name: str
+    address: str
```
daejunpark, line 10: naming: precisely speaking, this is not an address, but a selector ([4-byte signature](https://www.4byte.directory/)) for functions, events, or errors.

clarifying this or renaming it would be great.

src/halmos/mapper.py
```diff
@@ -0,0 +1,181 @@
...
+                node for node in contract_info.nodes if node.address == address
+            ]
+
+            for node in matching_nodes:
+                result += f"{key}.{node.name} "
+
+        return result.strip() if result != "" and address != "0x" else address
+
+
+class DeployAddressMapper(metaclass=SingletonMeta):
```
daejunpark, line 161: i'm concerned about DeployAddressMapper being a singleton object. unlike `Mapper` which is per project, `DeployAddressMapper` is per test. sharing a singleton across multiple tests don't seem like a good practice to me. also, i don't think this would would work with --test-parallel, due to potential race conditions. what do you think? cc: @karmacoma-eth

daejunpark (commented):
amazing, thanks for the contribution! i left a couple of comments.

src/halmos/mapper.py
```diff
@@ -0,0 +1,181 @@
...
+    _instances: Dict[Type, Any] = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+
+        return cls._instances[cls]
+
+
+class Mapper(metaclass=SingletonMeta):
```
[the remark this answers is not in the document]
karmacoma-eth: I think this makes sense, it is logically a global object and it is immutable-like (after the parsing phase). I'm not too worried about enforcing the immutability here, a comment should be fine

src/halmos/mapper.py
karmacoma-eth: good catch, I don't think DeployAddressMapper should be a singleton (or it should be scoped to a single test context). 

It seems to currently work with `--test-parallel` because it is not updated during the test run, it is only updated lazily in `render_trace` (so from the main thread). This likely means that values from a test can "leak" to another thread

karmacoma-eth: little repro:

```solidity
contract A {}

contract Test82 is Test, SymTest {
    function setUp() public {
        console2.log("Test82.setUp");
    }

    function test_DeployMapperLeak1_double_deploy() external {
        A a = new A();
        assert(address(a) != address(b));
    }

    function test_DeployMapperLeak2_deploys_nothing() external {
        vm.deal(address(this), 1 ether);
        payable(address(0xaaaa0002)).transfer(1 ether);
    }
}
```

if `deploys_nothing` runs first, it has the expected trace (`0xaaaa0002` is a naked address, not an A)

```
Trace:
    CALL Test82::test_DeployMapperLeak2_deploys_nothing
        CALL HEVM_ADDRESS::0xc88a5e6d00000000000000000000000000000000000000000000000000000000aaaa00010000000000000000000000000000000000000000000000000de0b6b3a7640000
        ↩ 0x
        CALL 0xaaaa0002::0x (value: 1000000000000000000)
        ↩ f_ret_256(call_exit_code_02())
    ↩ REVERT f_ret_256(call_exit_code_02()) (error: Revert())
```

If `double_deploy` runs first, the `DeployAddressMapper` is now polluted and the trace for `deploys_nothing` wrongly indicates that eth was transferred to an A instance:

```
Trace:
    CALL Test82::test_DeployMapperLeak2_deploys_nothing
        CALL HEVM_ADDRESS::0xc88a5e6d00000000000000000000000000000000000000000000000000000000aaaa00010000000000000000000000000000000000000000000000000de0b6b3a7640000
        ↩ 0x
        CALL A::0x (value: 1000000000000000000)    # wrong label here
        ↩ f_ret_256(call_exit_code_02())
    ↩ STOP 0x
```

daejunpark: thanks for further clarification, karma!

if it is nontrivial to fix this, i'm ok with adding a todo comment, and opening a separate pr to fix it later.

karmacoma-eth: indeed, let's merge as-is and fix up later. I wanted to try a quick fix by introducing scopes for the singletons, but it's trickier than I thought

karmacoma-eth (approved):

daejunpark (approved):
left todo notes for further improvements

PR #310, opened 2024-06-23
refactor: data structure of path conditions

daejunpark:


src/halmos/sevm.py
```diff
@@ -477,28 +477,34 @@ class Path:
...
-    branching: List  # indexes of conditions
+    conditions: Dict  # cond -> bool (true if explicit branching conditions)
     pending: List
 
     def __init__(self, solver: Solver):
         self.solver = solver
         self.num_scopes = 0
-        self.conditions = []
-        self.branching = []
+        self.conditions = {}
```
karmacoma-eth, line 486: neat!

src/halmos/sevm.py
```diff
@@ -477,28 +477,34 @@ class Path:
...
         raise NotImplementedError(f"use the branch() method instead of deepcopy()")
 
     def __str__(self) -> str:
-        branching_conds = [self.conditions[idx] for idx in self.branching]
         return "".join(
-            [f"- {cond}\n" for cond in branching_conds if str(cond) != "True"]
+            [
+                f"- {cond}\n"
+                for cond in self.conditions
+                if self.conditions[cond] and str(cond) != "True"
```
karmacoma-eth, line 498: `str(cond)` can be expensive, we should use `is_true(cond)` here

PR #311, opened 2024-06-24
feat: cache unsat queries using unsat core

daejunpark:
this pr implements a solver caching feature by utilizing the unsat core.

the key idea: when a query contains a known unsat core, it is also unsat, thus no need to run the smt solver.

when --cache-solver is enabled, the smt solver is asked to compute the unsat core for unsat queries.  subsequent queries are then compared with the previously computed unsat cores before being sent to the solver, allowing us to skip solving if they have already been known as unsat.

caching is particularly beneficial when there are many similar paths. since similar paths may share the same unsat core, they can be discharged after only a single solving, rather than repeating the same solving process.

for example, in the snekmate erc721 test, the runtime reduces from 477.48s to 295.95s:
- without caching:
```
# of potential paths involving assertion violations: 130 / 238  (--solver-threads 8)
[PASS] testHalmosAssertNoBackdoor(...) (paths: 238, time: 477.48s (paths: 166.15s, models: 311.33s), bounds: [])
```
- with caching: 
```
# of potential paths involving assertion violations: 127 / 234  (--solver-threads 8)
[PASS] testHalmosAssertNoBackdoor(...) (paths: 234, time: 295.95s (paths: 208.52s, models: 87.43s), bounds: [])
```

note: computing unsat cores incurs a performance overhead, so caching may not be beneficial for small tests.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -1079,13 +1079,15 @@ def solve(
         with open(dump_filename, "w") as f:
             if args.verbose >= 1:
                 print(f"Writing SMT query to {dump_filename}")
-            f.write("(set-option :produce-unsat-cores true)\n")
+            if args.cache_solver:
+                f.write("(set-option :produce-unsat-cores true)\n")
             f.write("(set-logic QF_AUFBV)\n")
             f.write(query.smtlib)
             f.write(named_assertions)
             f.write("(check-sat)\n")
             f.write("(get-model)\n")
-            f.write("(get-unsat-core)\n")
+            if args.cache_solver:
+                f.write("(get-unsat-core)\n")
 
     if args.solver_command:
         if args.verbose >= 1:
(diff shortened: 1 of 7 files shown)


src/halmos/__main__.py
```diff
@@ -1035,27 +1040,62 @@ def run_sequential(run_args: RunArgs) -> List[TestResult]:
...
+                    f"(assert (! |{assert_id}| :named <{assert_id}>))\n"
+                    for assert_id in query.assertions
+                ]
+            )
+
         with open(dump_filename, "w") as f:
             if args.verbose >= 1:
                 print(f"Writing SMT query to {dump_filename}")
+            if args.cache_solver:
+                f.write("(set-option :produce-unsat-cores true)\n")
```
karmacoma-eth, line 1090: is that a z3 specific option?

karmacoma-eth: TIL: no it's not!

karmacoma-eth:
this is actually very impressive, hot damn

PR #322, opened 2024-06-29
perf: early exit for fail()

daejunpark:
fix fail() to exit early.

previously, fail() was treated as a normal evm exception, allowing execution to continue after returning to the caller if any. (see the new test as example.) such continued execution is unnecessary. now fail() exits immediately, avoiding potential performance degradation.

### DIFF
--- a/src/halmos/exceptions.py
+++ b/src/halmos/exceptions.py
@@ -8,11 +8,22 @@ Note: this is a modified version execution-specs' src/ethereum/<fork>/vm/excepti
 """
 
 
-class InfeasiblePath(Exception):
+class PathEndingException(Exception):
+    """
+    Base class for any exception that should stop the current path exploration.
+
+    Stopping path exploration means stopping not only the current EVM context but also its parent contexts if any.
+    """
+
     pass
 
 
-class HalmosException(Exception):
+class HalmosException(PathEndingException):
+    """
+    Base class for unexpected internal errors happening during a test run.
+    Inherits from RunEndingException because it should stop further path exploration.
+    """
+
     pass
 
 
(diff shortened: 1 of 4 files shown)


src/halmos/exceptions.py
```diff
@@ -20,6 +20,22 @@ class NotConcreteError(HalmosException):
     pass
 
 
+class HalmosTargetReached(Exception):
```
daejunpark, line 23: please feel free to suggest a better name

karmacoma-eth: it sounds a little too semantically charged for me, I would go for something more descriptive like `RunEndingException(Exception)`, and I would make both `HalmosException` and `FailCheatcode` inherit from `RunEndingException`:

```py
class RunEndingException(Exception):
    """
    Base class for any exception that should not just stop the current EVM context, instead aborting the current test run. 
    """

    pass


class HalmosException(RunEndingException):
    """
    Base class for unexpected internal errors happening during a test run. 
    Inherits from RunEndingException because it should not just stop the current EVM context, but stop the whole run. 
    """

    pass


class FailCheatcode(RunEndingException):
    """
    Raised when invoking DSTest's fail() pseudo-cheatcode.
    Inherits from RunEndingException because it should not just stop the current EVM context, but stop the whole run. 
    """

    pass
```

karmacoma-eth: some alternatives along the same idea:

- AbortCallException 
- AbortTestException
- RunHaltException
- TestStopException

tests/regression/test/Foundry.t.sol
```diff
@@ -16,11 +16,27 @@ contract DeepFailer is Test {
...
         DeepFailer(address(this)).do_test(0);
     }
 }
 
+contract EarlyFailTest is Test {
+    function do_fail() external {
+        fail();
+    }
+
+    function check_early_fail_cheatcode(uint x) public {
```
karmacoma-eth, line 29: nice test 👍

tests/regression/test/Foundry.t.sol
```diff
@@ -16,11 +16,27 @@ contract DeepFailer is Test {
...
+        address(this).call(abi.encodeWithSelector(this.do_fail.selector, ""));
```
karmacoma-eth, line 30:
```suggestion
        # we want `fail()` to happen in a nested context,
        # to test that it ends not just the current context but the whole run
        address(this).call(abi.encodeWithSelector(this.do_fail.selector, ""));
```

karmacoma-eth (approved):

src/halmos/exceptions.py
daejunpark: thanks for suggestion! i ended up with `PathEndingException`, because "run" is ambiguous with the entire test run.

karmacoma-eth: nice, love it ♥️

PR #306, opened 2024-06-06
feat: Support assert cheatcodes for forge-std 1.8.0

EmperorOrokuSaki:
Work-in-progress
Related issue: #256

PR #323, opened 2024-06-29
feat: support for forge-std assert cheatcodes

daejunpark:
based on: #306

not yet supported: (to be added in a separate pr)
- assertEq for bytes[]
- assertEqDecimal

won't support: (symbolically approximate comparison is too challenging)
- assertApproxEq


karmacoma-eth:
amazing work, thanks @EmperorOrokuSaki and @daejunpark !

PR #329, opened 2024-07-23
fix bad abi encoding in createBytes4

karmacoma-eth:
bytes4 must be right-padded with 0s, not left padded


daejunpark (approved):
great catch!

PR #330, opened 2024-07-24
fix workflows

karmacoma-eth:


.gitmodules
```diff
@@ -1,18 +1,24 @@
...
 	url = https://github.com/transmissions11/solmate
+	shallow = true
 [submodule "tests/lib/solady"]
 	path = tests/lib/solady
 	url = https://github.com/Vectorized/solady
+	shallow = true
 [submodule "tests/lib/multicaller"]
 	path = tests/lib/multicaller
 	url = https://github.com/Vectorized/multicaller
+	shallow = true
```
daejunpark, line 24: what's the effect of this in the normal dev setting for cloning and updating submodules? is this better than having the ci script to use --depth 1 when cloning?

daejunpark (approved):

.gitmodules
karmacoma-eth: not clear, didn't seem to move metrics. It's also weirdly hard to test, because I can't check out this commit directly. I'm not convinced this does anything 😫

Issue #281, opened 2024-04-28
Remove `metadata` checks for files with free functions / file-level functions


pcaversaccio:
**Describe the bug**

If you currently use in some form contracts that use [free functions](https://soliditylang.org/blog/2020/09/02/solidity-0.7.1-release-announcement/) (a.k.a file-level functions) via imports, `halmos` will issue the following warning:

```sh
Skipped FreeFunctionsFile.json due to parsing failure: KeyError: 'metadata'
```

Currently, this warning is printed in red colour. I would suggest orange should be enough ;). The underlying reason is that these files are compiled in the normal way (since in some contracts you will use them) but there is no `metadata` available in the standard JSON input file for these files.

**To Reproduce**

`git` clone `https://github.com/pcaversaccio/snekmate` and checkout the `feat/halmos` branch. Run `pnpm install` and `forge install` and invoke:

```sh
halmos --contract ERC20TestHalmos --function testHalmosAssertNoBackdoor --storage-layout generic --ffi -vvvv
```

![image](https://github.com/a16z/halmos/assets/25297591/8aba1ebc-ffdc-4203-a7e6-4b26cefc22de)

**Environment:**
 - OS: `Linux`
 - Python version: `3.11.3`
 - Halmos and other dependency versions: `0.1.13.dev1+ga617814`

karmacoma-eth:
Trying to get to the bottom of this issue:

using forge 0.2.0 (9444c62 2024-07-26T00:22:04.784959000Z) and solc 0.8.26 (so both up to date), halmos invokes `forge build` with the `--extra-output metadata` flag. 

We use this to parse the compiler version in each `out/<filename>.sol/<contract_or_filename>.json`.

However, when a solidity does not contain a contract body, it seems that the `metadata` section of the json output is missing. Some examples where this can happen:

console2.sol, just an import:

```
// SPDX-License-Identifier: MIT
pragma solidity >=0.4.22 <0.9.0;

import {console as console2} from "./console.sol";
```

Just a sol file with no code at all:

```
// SPDX-License-Identifier: MIT
pragma solidity >=0.4.22 <0.9.0;
```

A sol file with free functions (like https://github.com/transmissions11/solmate/blob/main/src/utils/SignedWadMath.sol):

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.26;

function inc(uint256 x) pure returns (uint256) {
    return x + 1;
}
```

[...]

PR #331, opened 2024-07-25
chore: only warn about KeyError "ast"/"metadata" if debugging

karmacoma-eth:
plus a drive-by simplification of Mapper


src/halmos/__main__.py
```diff
@@ -1357,30 +1366,19 @@ def parse_build_out(args: HalmosConfig) -> Dict:
...
-                            nodes=[],
+                contract_map[contract_name] = (json_out, contract_type, natspec)
+                parse_symbols(args, contract_map, contract_name)
+
+            except KeyError as err:
+                if str(err) in ["ast", "metadata"]:
+                    # somewhat expected, e.g. for sol files containing only libraries
+                    if args.debug:
+                        debug(
+                            f"Skipped {json_filename} due to parsing failure: "
```
karmacoma-eth, line 1378: this is the real change, the rest is a refactoring

src/halmos/__main__.py
```diff
@@ -1357,30 +1366,19 @@ def parse_build_out(args: HalmosConfig) -> Dict:
...
-                    else:
-                        contract_mapping_info.bytecode = bytecode
-
-                    contract_mapping_info = Mapper().get_contract_mapping_info_by_name(
-                        contract_name
-                    )
-                    Mapper().parse_ast(contract_map[contract_name][0]["ast"])
 
-                except Exception:
-                    pass
```
karmacoma-eth, line 1383: this is what prompted the refactoring, we don't want to ignore all exceptions

src/halmos/__main__.py
```diff
@@ -1300,8 +1296,24 @@ def mk_arrlen(args: HalmosConfig) -> Dict[str, int]:
     return arrlen
 
 
+def parse_symbols(args: HalmosConfig, contract_map, contract_name):
```
karmacoma-eth, line 1299: pulling this in its own method to avoid nested try..except blocks

src/halmos/__main__.py
```diff
@@ -1300,8 +1296,24 @@ def mk_arrlen(args: HalmosConfig) -> Dict[str, int]:
...
+    except Exception:
+        if args.debug:
+            debug(f"error parsing symbols for contract {contract_name}")
+            debug(traceback.format_exc())
+
+
 def parse_build_out(args: HalmosConfig) -> Dict:
-    result = {}  # compiler version -> source filename -> contract name -> (json, type)
+    # compiler version -> source filename -> contract name -> (json, type)
+    result = defaultdict(lambda: defaultdict(dict))
```
karmacoma-eth, line 1316: opportunistic, this avoids doing `if x not in y: y[x] = {}` a bunch of times

src/halmos/__main__.py
```diff
@@ -1318,21 +1330,17 @@ def parse_build_out(args: HalmosConfig) -> Dict:
...
+                continue
+            if json_filename.startswith("."):
+                continue
 
-                json_path = os.path.join(sol_path, json_filename)
-                with open(json_path, encoding="utf8") as f:
-                    json_out = json.load(f)
+            json_path = os.path.join(sol_path, json_filename)
+            with open(json_path, encoding="utf8") as f:
+                json_out = json.load(f)
```
karmacoma-eth, line 1340: pulling this out of the try..except block to reduce the scope of the try blocks and minimize nesting

karmacoma-eth:
should address https://github.com/a16z/halmos/issues/281

src/halmos/__main__.py
daejunpark: what's your intention when err is not ast nor metadata? generate warning, or silently ignore?

daejunpark:
could you separate the main change from the refactoring changes, to make it easier to review the core change? could be a separate pr, or at least a different commit.

src/halmos/__main__.py
daejunpark: another q: would it be possible to check if the ast / metadata missing error is indeed due to libraries?

karmacoma-eth: > another q: would it be possible to check if the ast / metadata missing error is indeed due to libraries?

that was the case every time I looked

karmacoma-eth: > what's your intention when err is not ast nor metadata? generate warning, or silently ignore?

oops, the intention is to generate a warning for other errors

daejunpark: > > what's your intention when err is not ast nor metadata? generate warning, or silently ignore?
> 
> oops, the intention is to generate a warning for other errors

in that case, i don't think the current code works as intended.

karmacoma-eth: fixed in b135251

daejunpark:
[starting a new thread for easier discussion]

> > another q: would it be possible to check if the ast / metadata missing error is indeed due to libraries?
> 
> that was the case every time I looked

ah, my question was, whether it's possible to automatically check if the problematic file is a library or not, and generate the error message only when it is not a library.

indeed, i'm concerned about silently ignoring field-missing errors in non-debug mode. although we haven't seen cases where non-library files are skipped due to missing fields, i believe we still need to allow users review the skipped files, unless we're certain they can be safely skipped.

also, the usability issue described in #281 is not about the warning message itself, but rather that the message is not detailed enough to understand what's happening, and its tone is too strong.

so, my suggestion for the short term is to improve the warning message, allowing users to review them and detect any real issues, without being scared.

a longer-term solution is to implement a mechanism to check whether the given file is a library and ignore the error only when it's a library.

karmacoma-eth:
closing this one, it's getting too messy. Will update the investigation in https://github.com/a16z/halmos/issues/281

PR #340, opened 2024-08-13
smt: abstract multiplication

daejunpark:
Abstract multiplication in a similar way to division and modulo operations.

The actual bitvector multiplication is used when either argument is concrete. Also, when an argument is a power of two, the left-shift is used instead of multiplication.

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -1356,10 +1356,13 @@ class SEVM:
             return w1 - w2
 
         if op == EVM.MUL:
-            if is_bv_value(w1) and is_bv_value(w2):
+            is_bv_value_w1 = is_bv_value(w1)
+            is_bv_value_w2 = is_bv_value(w2)
+
+            if is_bv_value_w1 and is_bv_value_w2:
                 return w1 * w2
 
-            if is_bv_value(w1):
+            if is_bv_value_w1:
                 i1: int = w1.as_long()
                 if i1 == 0:
                     return w1
@@ -1370,7 +1373,7 @@ class SEVM:
                 if is_power_of_two(i1):
                     return w2 << (i1.bit_length() - 1)
 
-            if is_bv_value(w2):
+            if is_bv_value_w2:
                 i2: int = w2.as_long()
                 if i2 == 0:
                     return w2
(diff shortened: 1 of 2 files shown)


src/halmos/sevm.py
```diff
@@ -1348,7 +1356,35 @@ def arith(self, ex: Exec, op: int, w1: Word, w2: Word) -> Word:
...
+            if is_bv_value(w1):
+                i1: int = w1.as_long()
+                if i1 == 0:
+                    return w1
+
+                if i1 == 1:
+                    return w2
+
+                if is_power_of_two(i1):
+                    return w2 << int(math.log(i1, 2))
```
karmacoma-eth, line 1371:
```suggestion
                    return w2 << i1.bit_length() - 1
```

quick check:

~/projects/halmos on code-fastpath wip *21 !2 ?50 ❯ python -m timeit --setup='import math' 'int(math.log(32, 2))'                                                                                                                                                                     
2000000 loops, best of 5: 121 nsec per loop

~/projects/halmos on code-fastpath wip *21 !2 ?50 ❯ python -m timeit 'int(32).bit_length() - 1'                                                                                                                                                                 
5000000 loops, best of 5: 66.1 nsec per loop

src/halmos/sevm.py
```diff
@@ -1348,7 +1356,35 @@ def arith(self, ex: Exec, op: int, w1: Word, w2: Word) -> Word:
             return w1 - w2
 
         if op == EVM.MUL:
-            return w1 * w2
+            if is_bv_value(w1) and is_bv_value(w2):
```
karmacoma-eth, line 1359: can we invoke `is_bv_value(x)` once for each and reuse the results? in the worst case it results in 6 calls to that function, which itself makes more function calls

src/halmos/sevm.py
```diff
@@ -1348,7 +1356,35 @@ def arith(self, ex: Exec, op: int, w1: Word, w2: Word) -> Word:
...
+                return w1 * w2
```
karmacoma-eth, line 1360: just to double check: we're fine here in case of overflow?

daejunpark: yes, and it's indeed intended.

src/halmos/sevm.py
daejunpark: done: https://github.com/a16z/halmos/pull/340/commits/2ee41207c8a111980670ff7f426fc20f1e57b89a

src/halmos/sevm.py
daejunpark: done: https://github.com/a16z/halmos/pull/340/commits/e8490ee2815662b77474a4bcbf5695ac1c119f4e

karmacoma-eth (approved):

PR #341, opened 2024-08-13
ci: use yices2 for test-external

daejunpark:
use yices2 for the external test, as bitwuzla in the docker image is unstable, causing the test to be flaky:
- https://github.com/a16z/halmos/actions/runs/10364288791/job/28689299861
- https://github.com/a16z/halmos/actions/runs/10103994098/job/27942255454
- https://github.com/a16z/halmos/actions/runs/10014074218/job/27683004640


karmacoma-eth (approved):
not sure if the flakiness comes from bitwuzla or docker or something else but happy to give this a try

PR #345, opened 2024-08-14
ci: disable pytest-xdist

daejunpark:
The ci test workflow was updated to use a single worker with `pytest -n 1` in [#336](https://github.com/a16z/halmos/pull/336/files#r1702170704). However, flakiness issues persist, so let's try disabling the xdist plugin altogether.

Additionally, xdist workers crash when --cache-solver enabled. Now that xdist is disabled, tests using --cache-solver have also been added.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,4 +53,4 @@ jobs:
         run: python -m pip install -e .
 
       - name: Run pytest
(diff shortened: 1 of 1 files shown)


karmacoma-eth:
ah really? darn

Do you remember what error they failed with?

anyway happy to give this a try

daejunpark:
> Do you remember what error they failed with?

i don't remember exactly, (though i could check the old gh action logs), but i believe it was segfault due to race conditions caused by multiple workers.

PR #349, opened 2024-08-20
feat: branching over symbolic call addresses

daejunpark:
When a call target address may be aliased with multiple existing addresses, or may be non-existent, explicitly branching into multiple paths for each potential alias or non-existent address.

Although this change may increase the number of potential paths, it eliminates the time needed to solve must-alias conditions, which can take up to 1s each. Also, the path branching is unlikely to exponentially grow, unless there are many duplicated instances of contracts with similar interfaces, because most paths will immediately revert due to unimplemented selectors. In such cases, the time saved by reducing solver overhead outweighs the cost of increased path branching.

As a result, "unknown calls" no longer exist, and the related options, `--uninterpreted-unknown-calls` and `--return-size-of-unknown-calls`, are no longer needed and have been deprecated.

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -1517,12 +1517,16 @@ class SEVM:
 
         self.storage_model.store(ex, addr, loc, val)
 
-    def resolve_address_alias(self, ex: Exec, target: Address, stack, step_id) -> Address:
+    def resolve_address_alias(
+        self, ex: Exec, target: Address, stack, step_id
+    ) -> Address:
         if target in ex.code:
             return target
 
         if self.options.debug:
-            debug(f"Address {hexify(target)} not in: [{", ".join([hexify(addr) for addr in ex.code])}]")
+            debug(
+                f"Address {hexify(target)} not in: [{', '.join([hexify(addr) for addr in ex.code])}]"
+            )
 
         if is_bv_value(target):
             if self.options.debug:
@@ -1537,10 +1541,12 @@ class SEVM:
             alias_cond = target == addr
             if ex.check(alias_cond) != unsat:
                 if self.options.debug:
-                    debug(f"Potential address alias: {hexify(addr)} for {hexify(target)}")
+                    debug(
+                        f"Potential address alias: {hexify(addr)} for {hexify(target)}"
+                    )
                 potential_aliases.append((addr, alias_cond))
 
-        emptyness_cond = And([ target != addr for addr in ex.code ])
+        emptyness_cond = And([target != addr for addr in ex.code])
         if ex.check(emptyness_cond) != unsat:
             if self.options.debug:
                 debug(f"Potential empty address: {hexify(target)}")
@@ -1587,12 +1593,7 @@ class SEVM:
         ex.balance_update(to, self.arith(ex, EVM.ADD, ex.balance_of(to), value))
 
     def call(
-        self,
-        ex: Exec,
-        op: int,
-        stack: List[Tuple[Exec, int]],
-        step_id: int,
-        to_addr
+        self, ex: Exec, op: int, stack: List[Tuple[Exec, int]], step_id: int, to_addr
     ) -> None:
         gas = ex.st.pop()
         to = uint160(ex.st.pop())
@@ -1759,7 +1760,9 @@ class SEVM:
                 exit_code = con(1)
                 modulus_size = int_of(extract_bytes(arg, 64, 32))
                 f_modexp = Function(
-                    f"f_modexp_{arg_size}_{modulus_size}", BitVecSorts[arg_size], BitVecSorts[modulus_size]
+                    f"f_modexp_{arg_size}_{modulus_size}",
+                    BitVecSorts[arg_size],
+                    BitVecSorts[modulus_size],
                 )
                 # TODO: empty returndata in error
                 ret = ByteVec(f_modexp(arg))
@@ -1767,17 +1770,13 @@ class SEVM:
             # ecadd
             elif eq(to, con_addr(6)):
                 exit_code = con(1)
-                f_ecadd = Function(
-                    f"f_ecadd", BitVecSorts[1024], BitVecSorts[512]
-                )
+                f_ecadd = Function(f"f_ecadd", BitVecSorts[1024], BitVecSorts[512])
                 ret = ByteVec(f_ecadd(arg))
 
             # ecmul
             elif eq(to, con_addr(7)):
                 exit_code = con(1)
-                f_ecmul = Function(
-                    f"f_ecmul", BitVecSorts[768], BitVecSorts[512]
-                )
+                f_ecmul = Function(f"f_ecmul", BitVecSorts[768], BitVecSorts[512])
                 ret = ByteVec(f_ecmul(arg))
 
             # ecpairing
@@ -1791,9 +1790,7 @@ class SEVM:
             # blake2f
             elif eq(to, con_addr(9)):
                 exit_code = con(1)
-                f_blake2f = Function(
-                    f"f_blake2f", BitVecSorts[1704], BitVecSorts[512]
-                )
+                f_blake2f = Function(f"f_blake2f", BitVecSorts[1704], BitVecSorts[512])
                 ret = ByteVec(f_blake2f(arg))
 
             # point_evaluation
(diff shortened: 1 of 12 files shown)


src/halmos/sevm.py
```diff
@@ -1536,26 +1517,54 @@ def sstore(self, ex: Exec, addr: Any, loc: Any, val: Any) -> None:
 
         self.storage_model.store(ex, addr, loc, val)
 
-    def resolve_address_alias(self, ex: Exec, target: Address) -> Address:
+    def resolve_address_alias(
+        self, ex: Exec, target: Address, stack, step_id
+    ) -> Address:
```
daejunpark, line 1522: i don't like passing the worklist items as additional arguments, but i tried to keep this pr minimal, and plan to refactor this in a future pr.

src/halmos/sevm.py
```diff
@@ -1774,13 +1731,76 @@ def call_unknown() -> None:
...
+                ret = ByteVec(f_blake2f(arg))
+
+            # point_evaluation
+            elif eq(to, con_addr(10)):
+                exit_code = con(1)
+                f_point_evaluation = Function(
+                    f"f_point_evaluation", BitVecSorts[1544], BitVecSorts[512]
+                )
+                ret = ByteVec(f_point_evaluation(arg))
+
```
daejunpark, line 1803: will refactor these precompile semantics by moving into a separate module

src/halmos/sevm.py
```diff
@@ -2338,22 +2344,32 @@ def finalize(ex: Exec):
...
+                    ex.st.pop()
+
                     if account_addr is not None:
                         codesize = len(ex.code[account_addr])
+                    elif (
+                        eq(account, hevm_cheat_code.address)
+                        or eq(account, halmos_cheat_code.address)
+                        or eq(account, console.address)
+                    ):
+                        codesize = 777  # dummy arbitrary value
```
daejunpark, line 2360: any suggestions for better values for the codesize of cheatcode contracts?

src/halmos/sevm.py
```diff
@@ -2338,22 +2344,32 @@ def finalize(ex: Exec):
...
                     else:
-                        codesize = f_extcodesize(account)
-                        if (
-                            eq(account, hevm_cheat_code.address)
-                            or eq(account, halmos_cheat_code.address)
-                            or eq(account, console.address)
-                        ):
-                            ex.path.append(codesize > 0)
+                        codesize = 0
+
```
daejunpark, line 2363: todo: double-check if the codesize of precompiles is indeed zero

src/halmos/sevm.py
```diff
@@ -2381,14 +2392,18 @@ def finalize(ex: Exec):
...
+                    )
+                    ex.st.pop()
+
                     addr = alias_addr if alias_addr is not None else account_addr
 
                     account_code: Optional[Contract] = ex.code.get(addr, None)
 
                     codehash = (
-                        f_extcodehash(addr)
+                        EMPTY_KECCAK
```
daejunpark, line 2406: is this correct?

karmacoma-eth:
tangential but related, do you think this test should pass?

```solidity
    function check_unknown_initial_balance(address addr) public {
        assertEq(addr.balance, 0);
    }
```

daejunpark:
> tangential but related, do you think this test should pass?
> 
> ```solidity
>     function check_unknown_initial_balance(address addr) public {
>         assertEq(addr.balance, 0);
>     }
> ```

eventually yes, but in a separate pr. this pr is only for handling code of symbolic addresses.

karmacoma-eth (approved):

src/halmos/sevm.py
```diff
@@ -2381,14 +2392,18 @@ def finalize(ex: Exec):
                         ex.st.memory.set_slice(loc, end_loc, codeslice)
 
                 elif opcode == EVM.EXTCODEHASH:
-                    account_addr = uint160(ex.st.pop())
-                    alias_addr = self.resolve_address_alias(ex, account_addr)
+                    account_addr = uint160(ex.st.peek())
+                    alias_addr = self.resolve_address_alias(
+                        ex, account_addr, stack, step_id
+                    )
+                    ex.st.pop()
```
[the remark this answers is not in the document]
daejunpark: yes :(

src/halmos/utils.py
```diff
@@ -184,6 +184,9 @@ def create_solver(logic="QF_AUFBV", ctx=None, timeout=0, max_memory=0):
     # QF_AUFBV: quantifier-free bitvector + array theory: https://smtlib.cs.uiowa.edu/logics.shtml
     solver = SolverFor(logic, ctx=ctx)
 
+    # reset any remaining solver states from the default context
+    solver.reset()
```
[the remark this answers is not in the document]
daejunpark: for the first question, yes, but mostly when a previous solver terminates abnormally, timeout, etc.

for the second question, no, we need to have some symbolic terms (e.g., zero, one) to be shared among different solvers used in different tests.

src/halmos/sevm.py
daejunpark: what? so, extcodehash of a contract account with empty code is different from that of a non-existing account?

so, here we should return 0, based on the current implementation of vms?

src/halmos/sevm.py
```diff
@@ -1584,11 +1593,7 @@ def transfer_value(
         ex.balance_update(to, self.arith(ex, EVM.ADD, ex.balance_of(to), value))
 
     def call(
-        self,
-        ex: Exec,
-        op: int,
-        stack: List[Tuple[Exec, int]],
-        step_id: int,
+        self, ex: Exec, op: int, stack: List[Tuple[Exec, int]], step_id: int, to_addr
```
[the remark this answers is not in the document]
daejunpark: yes, you're right, and i agree

src/halmos/sevm.py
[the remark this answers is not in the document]
karmacoma-eth: alright well if it works it works, but would be nice to find a better pattern if we can

src/halmos/sevm.py
karmacoma-eth: > what? so, extcodehash of a contract account with empty code is different from that of a non-existing account?

yes, https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1052.md#specification is a bit more explicit

>>> The EXTCODEHASH of the account without code is c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470 what is the keccack256 hash of empty data.
The EXTCODEHASH of non-existent account is 0.
The EXTCODEHASH of a precompiled contract is either c5d246... or 0.

karmacoma-eth: > so, here we should return 0, based on the current implementation of vms?

we need to have a more proper `is_empty` check for the account, which is also affected by the balance. This needs to pass:

```solidity
    function testAccountWithOrWithoutBalance() public {
        address a = address(0x42);
        assertEq(extcodehash(a), bytes32(0), "empty account has extcodehash 0");

        vm.deal(a, 1 ether);
        assertEq(extcodehash(a), 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470, "account with balance has extcodehash keccak('')");
    }
```

karmacoma-eth: full test:

```solidity
pragma solidity ^0.8.18;

import "forge-std/Test.sol";

contract Empty {
    constructor() {
        assembly {
            return(0, 0)
        }
    }
}

contract EmptyDeployerTest is Test {
    function extcodehash(address _addr) public view returns (bytes32) {
        bytes32 hash;
        assembly {
            hash := extcodehash(_addr)
        }
        return hash;
    }

    function testEmptyDeployer() public {
        address a = vm.computeCreateAddress(address(this), vm.getNonce(address(this)));
        assertEq(extcodehash(a), bytes32(0), "empty account has extcodehash 0");
        assertEq(vm.getNonce(a), 0);

        // do deploy
        new Empty();

        assertEq(a.code.length, 0);
        assertEq(extcodehash(a), keccak256(""));
        assertEq(vm.getNonce(a), 1);
    }


    function testAccountWithOrWithoutBalance() public {
        address a = address(0x42);
        assertEq(extcodehash(a), bytes32(0), "empty account has extcodehash 0");

        vm.deal(a, 1 ether);
        assertEq(extcodehash(a), keccak256(""), "account with balance has extcodehash keccak('')");
    }
}
```

in our case if we don't want to model nonces, we can probably differentiate between not being in `code`, vs being in `code` with empty bytecode

src/halmos/sevm.py
daejunpark: done

src/halmos/sevm.py

PR #350, opened 2024-08-23
deprecate --bytecode and --reset-bytecode

daejunpark:
deprecate two unused bytecode features to reduce code maintenance costs.


karmacoma-eth (approved):
nice ✂️✂️✂️

PR #351, opened 2024-08-23
feat: reject multiple setup paths

daejunpark:
currently, if multiple setup paths exist, one of them is randomly selected and used for testing. since the existence of multiple setup paths are likely due to an incorrect setup(), this behavior makes debugging more challenging. to address this, now it will fail immediately when multiple setup paths are detected.

to be merged after #350


tests/regression/test/Context.t.sol
```diff
@@ -191,7 +191,10 @@ contract ContextTest is Test {
     }
 
     function check_call0_normal(uint mode0) public payable {
-        require(mode0 < 9);
+        // vm.assume(mode0 < 9);
+        // NOTE: explicitly branch over mode0, as an infeasible path with mode0 >= 9 may not be eliminated due to an extremely inefficient solver environment (e.g, github workflow)
+        mode0 = split_mode_up_to_9(mode0);
+
         _check_call0(mode0);
```
karmacoma-eth, line 198: wait what? I don't understand what's going on here

karmacoma-eth (approved):
I agree with the general direction, however:
- we still have quite a few symbolic sources like balances, right? we might want to tighten these as much as possible 
- do we want this to be a hard break or more of a nudge? i.e. maybe allow a bypass with a CLI flag?

tests/regression/test/Context.t.sol
daejunpark: this is not related to this pr. it's a temporary fix to a flaky test that i found.

details of the flaky test: when mode0 == 9, `_check_call0()` fails due to HalmosException (which is intentional). however, the goal of this test is to consider only success paths, so the assumption `mode0 < 9` is provided at the beginning. however, the failed path may still appear after path exploration, if the branching solver couldn't solve the infeasible path condition of the failed path. this won't happen under normal circumstances, as the path condition `mode0 < 9 && mode0 == 9` is obviously unsat. but in the github runner, the solver may timeout, leading to a test failure, due to the failed path, even if it is infeasible. a proper fix is more involved, and will be implemented later.

karmacoma-eth: got it, thank you

daejunpark:
> * we still have quite a few symbolic sources like balances, right? we might want to tighten these as much as possible

balance is addressed in #352 

> * do we want this to be a hard break or more of a nudge? i.e. maybe allow a bypass with a CLI flag?

we could add a bypass flag, or make this optional (e.g., --strict), later if this turns out to cause test failures too often.

PR #352, opened 2024-08-23
feat: empty initial network state

daejunpark:
changes:
| | old | new |
| - | - | - |
| *.balance | symbolic | zero |
| msg.sender | 0xaaaa0000 | foundry caller address |
| tx.origin | symbolic | foundry caller address |
| this | 0xaaaa0001 | foundry test contract address |
| block.* | symbolic | foundry default |


src/halmos/__main__.py
```diff
@@ -221,41 +223,29 @@ def mk_calldata(
...
-def mk_caller(args: HalmosConfig) -> Address:
-    return mk_addr("msg_sender") if args.symbolic_msg_sender else magic_address
-
-
 def mk_this() -> Address:
     # NOTE: Do NOT remove the `con_addr()` wrapper.
     #       The return type should be BitVecSort(160) as it is used as a key for ex.code.
     #       The keys of ex.code are compared using structural equality with other BitVecRef addresses.
-    return con_addr(magic_address + 1)
+    return con_addr(FOUNDRY_TEST)
```
karmacoma-eth, line 248: might be nice to standardize on `address()` rather than `con_addr()`, wdyt?

src/halmos/cheatcodes.py
```diff
@@ -687,7 +690,7 @@ def handle(sevm, ex, arg: ByteVec, stack, step_id) -> ByteVec | None:
 
             return ret
 
-        # getBlockNumber() return (uint256)
+        # vm.getBlockNumber() return (uint256)
         elif funsig == hevm_cheat_code.get_block_nunber_sig:
```
karmacoma-eth, line 694: unrelated, but typo here

```suggestion
        elif funsig == hevm_cheat_code.get_block_number_sig:
```

karmacoma-eth: (needs to be fixed at definition site too)

src/halmos/config.py
```diff
@@ -165,16 +165,6 @@ class Config:
...
-    symbolic_msg_sender: bool = arg(
-        help="set msg.sender symbolic",
-        global_default=False,
-    )
-
-    no_test_constructor: bool = arg(
-        help="do not run the constructor of test contracts",
-        global_default=False,
-    )
-
```
karmacoma-eth, line 177: we're really cleaning up 🤗

tests/regression/test/Deal.t.sol
```diff
@@ -6,6 +6,10 @@ import "forge-std/Test.sol";
 contract DealTest is Test {
     C c;
 
+    function check_initial_balance(address addr) public {
+        assertEq(addr.balance, 0);
```
karmacoma-eth, line 10: 👍

karmacoma-eth (approved):

src/halmos/cheatcodes.py
daejunpark: good catch, fixed, thanks!

src/halmos/__main__.py
daejunpark: good idea. they are slightly different though right now--value range check only in con_addr. perhaps it would be better to add the value range check in the con(), and then replace con_addr() with address(). but we need to make sure this doesn't break others. let's work on it in a separate pr.

PR #354, opened 2024-08-26
add ruff to pre-commit and replace black action with pre-commit

karmacoma-eth:


pyproject.toml
```diff
@@ -54,3 +54,6 @@ select = [
 ignore = [
     "E501",  # line too long
 ]
+exclude = [
+    "tests/lib/**"
```
karmacoma-eth, line 58: we don't want to run ruff on solady or vyper stuff we may import in external libs

src/halmos/sevm.py
```diff
@@ -414,6 +415,8 @@ def push(self, v: Word) -> None:
             self.stack.append(simplify(v))
 
     def pop(self) -> Word:
+        if not self.stack:
+            raise StackUnderflowError()
```
karmacoma-eth, line 419: this used to fail with the raw Python `IndexError`, and caught by `pytest.raises(Exception)`

ruff rightfully points out that catching `Exception` is not great, so we can finally address the todo and wrap the exception in the semantically correct `StackUnderflowError`

tests/conftest.py
```diff
@@ -1,3 +1,8 @@
+# workaround for ruff removing fixture imports
+# they look unused because of dependency injection by pytest
+from test_fixtures import *  # noqa
```
karmacoma-eth, line 3: a bit annoying to work around, relevant issue: https://github.com/astral-sh/ruff/issues/3295

tests/test_cli.py
```diff
@@ -76,16 +78,17 @@ def test_decode_mixed_bytecode():
     assert contract[0] == EVM.PUSH20
     assert contract[27] == EVM.RETURN
     assert contract[28] == EVM.STOP  # past the end
-
-    contract.valid_jump_destinations() == set()
+    assert contract.valid_jump_destinations() == set()
```
karmacoma-eth, line 81: ruff pointed out that we were missing the assert 😅

tests/test_config.py
```diff
@@ -35,7 +39,7 @@ def test_fresh_config_has_only_None_values():
 
 
 def test_default_config_immutable(config):
-    with pytest.raises(Exception):
+    with pytest.raises(dataclasses.FrozenInstanceError):
```
karmacoma-eth, line 42: more specific exception type 👌

daejunpark (approved):

PR #355, opened 2024-08-26
feat: replace --symbolic-storage with enableSymbolicStorage(address) cheatcode

daejunpark:
- deprecate --symbolic-storage, which enables symbolic storage for all accounts.
- introduce a new cheatcode that enables symbolic storage on a per-account basis.

this simplifies the code logic, and improves usability as well. note that the symbolic storage feature is not frequently used.

----

(to be merged after #352) todo:
- [x] rebase on main
- [x] merge the new halmos cheatcode pr: https://github.com/a16z/halmos-cheatcodes/pull/7
- [x] update the halmos-cheatcode submodule


src/halmos/sevm.py
```diff
@@ -755,10 +757,16 @@ def extend_path(self, path):
...
+class StorageData:
+    symbolic: bool = False
+    mapping: dict = field(default_factory=dict)
+
+
 class Exec:  # an execution path
     # network
     code: dict[Address, Contract]
-    storage: dict[Address, dict[int, Any]]  # address -> { storage slot -> value }
+    storage: dict[Address, StorageData]  # address -> { storage slot -> value }
```
karmacoma-eth, line 769: would a `defaultdict` work here? we wouldn't have to check if keys exist and explicitly initialize them

src/halmos/sevm.py
```diff
@@ -1138,23 +1144,23 @@ def init(cls, ex: Exec, addr: Any, slot: int, keys: tuple) -> None:
...
                         label, BitVecSort256
                     )
                 else:
-                    ex.storage[addr][slot][num_keys][size_keys] = ZERO
+                    ex.storage[addr].mapping[slot][num_keys][size_keys] = ZERO
             else:
                 # do not use z3 const array `K(BitVecSort(size_keys), ZERO)` when not ex.symbolic
                 # instead use normal smt array, and generate emptyness axiom; see load()
-                ex.storage[addr][slot][num_keys][size_keys] = cls.empty(
+                ex.storage[addr].mapping[slot][num_keys][size_keys] = cls.empty(
```
karmacoma-eth, line 1163: take this as a soft suggestion and please double check if it is equivalent. I just applied a couple of my heuristics:
- `setdefault` instead of `if x not in y: y[x] = ...`
- avoid repeated dict lookups, factor them out in their own variable (e.g. `storage = ex.storage[addr]`)
- reduce nesting with early exits when applicable (e.g. `if size_keys in slot_mapping[num_keys]: return`)
- move variable definitions close to their point of use

Gives the following:

```py
    @classmethod
    def init(cls, ex: Exec, addr: Any, slot: int, keys: tuple) -> None:
        assert_address(addr)
        storage = ex.storage[addr]
        storage.mapping.setdefault(slot, {})

        num_keys = len(keys)
        slot_mapping = storage.mapping[slot]
        slot_mapping.setdefault(num_keys, {})

        size_keys = cls.bitsize(keys)
        if size_keys in slot_mapping[num_keys]:
            return

        if size_keys != 0:
            # do not use z3 const array `K(BitVecSort(size_keys), ZERO)` when not ex.symbolic
            # instead use normal smt array, and generate emptyness axiom; see load()
            slot_mapping[num_keys][size_keys] = cls.empty(addr, slot, keys)
            return

        # size_keys == 0
        if storage.symbolic:
            label = f"storage_{id_str(addr)}_{slot}_{num_keys}_{size_keys}_00"
            slot_mapping[num_keys][size_keys] = BitVec(label, BitVecSort256)
        else:
            slot_mapping[num_keys][size_keys] = ZERO
```

src/halmos/sevm.py
```diff
@@ -1168,15 +1174,18 @@ def load(cls, ex: Exec, addr: Any, loc: Word) -> Word:
...
                 ex.path.append(
                     Select(cls.empty(addr, slot, keys), concat(keys)) == ZERO
                 )
             return ex.select(
-                ex.storage[addr][slot][num_keys][size_keys], concat(keys), ex.storages
+                ex.storage[addr].mapping[slot][num_keys][size_keys],
+                concat(keys),
+                ex.storages,
+                ex.storage[addr].symbolic,
             )
```
karmacoma-eth, line 1189: I would avoid the `else` block after the return and decrease nesting here

karmacoma-eth (approved):
left code style suggestions, but nothing blocking. Love where this is headed!

src/halmos/sevm.py
karmacoma-eth: taking it a couple steps further:

- factoring out `slot_mapping[num_keys]` 
- using the return value of setdefault)
- turn if else blocks into a single expression

becomes:

```python
    @classmethod
    def init(cls, ex: Exec, addr: Any, slot: int, keys: tuple) -> None:
        assert_address(addr)
        storage = ex.storage[addr]
        slot_mapping = storage.mapping.setdefault(slot, {})

        num_keys = len(keys)
        slot_mapping_num_keys = slot_mapping.setdefault(num_keys, {})

        size_keys = cls.bitsize(keys)
        if size_keys in slot_mapping_num_keys:
            return

        if size_keys != 0:
            # do not use z3 const array `K(BitVecSort(size_keys), ZERO)` when not ex.symbolic
            # instead use normal smt array, and generate emptyness axiom; see load()
            slot_mapping_num_keys[size_keys] = cls.empty(addr, slot, keys)
            return

        # size_keys == 0
        slot_mapping_num_keys[size_keys] = (
            mk_uint256(f"storage_{id_str(addr)}_{slot}_{num_keys}_{size_keys}_00")
            if storage.symbolic
            else ZERO
        )
```

karmacoma-eth: could be improved with a better name for `slot_mapping_num_keys`, but I don't have a good intuition for what it should be

daejunpark:
thanks for the refactoring suggestion! let me implement it in a separate pr, as this pr is already quite nontrivial.

src/halmos/sevm.py
```diff
@@ -1279,17 +1288,22 @@ def empty(cls, addr: BitVecRef, loc: BitVecRef) -> ArrayRef:
     @classmethod
     def init(cls, ex: Exec, addr: Any, loc: BitVecRef) -> None:
         assert_address(addr)
-        if loc.size() not in ex.storage[addr]:
-            ex.storage[addr][loc.size()] = cls.empty(addr, loc)
+        if loc.size() not in ex.storage[addr].mapping:
+            ex.storage[addr].mapping[loc.size()] = cls.empty(addr, loc)
```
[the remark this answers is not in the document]
daejunpark: how to use defaultdict when the default value is not a constant but a function of `loc`?

src/halmos/sevm.py
daejunpark: yes, done in #356

src/halmos/sevm.py
daejunpark: done in #356

src/halmos/sevm.py
daejunpark: done in #356

daejunpark:
> thanks for the refactoring suggestion! let me implement it in a separate pr, as this pr is already quite nontrivial.

done in #356

src/halmos/sevm.py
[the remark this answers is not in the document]
karmacoma-eth: > If default_factory is not None, it is called without arguments to provide a default value for the given key, this value is inserted in the dictionary for the key, and returned.

well that's a problem, I thought the key was passed to the factory 🤨 oh well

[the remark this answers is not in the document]
karmacoma-eth: tested this out of curiosity, it seems to work the way I want:

```python
class betterdict(dict):
    def __init__(self, factory):
        self.factory = factory
    def __missing__(self, key):
        val = self.factory(key)
        self[key] = val
        return val
```

I wonder if there's some weird downside to it that would explain why defaultdict doesn't behave this way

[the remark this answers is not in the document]
daejunpark: yeah, i will let this particular thing stay as is. this two-liner doesn't look too bad. many other parts have been improved in #356, which looks much better. this part of code is one of the most complicated/convoluted in the codebase, so i'd prefer not to use aggressive optimizations here. :sweat_smile:

PR #356, opened 2024-08-27
refactor: use defaultdict for storage model

daejunpark:
follow up from https://github.com/a16z/halmos/pull/355#issuecomment-2311332440


karmacoma-eth (approved):
appreciate the extra docs 👌

PR #359, opened 2024-08-27
fix failed() cheatcode error

### LINKED ISSUE #358: calling ds-test/test.sol's `failed()` results in 'NoneType' object has no attribute 'size'
**Describe the bug**

with halmos 0.1.14.dev15+g060a6d8.d20240813:

```solidity
pragma solidity ^0.8.15;

import "forge-std/Test.sol";

contract Test91 is Test {
    function test_dstestFailed() external {
        console2.log("failed() =", failed());
    }
}
```

```shell
halmos --function test_dstestFailed --debug

Executing test_dstestFailed
Address 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d not in: [0x7fa9385be102ac3eac297483dd6233d62b3e1496]
Empty address: 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d
Address 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d not in: [0x7fa9385be102ac3eac297483dd6233d62b3e1496]
Empty address: 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d
Address 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d not in: [0x7fa9385be102ac3eac297483dd6233d62b3e1496]
Empty address: 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d
[ERROR] test_dstestFailed()
AttributeError: 'NoneType' object has no attribute 'size'
  File "/Users/karma/projects/halmos/src/halmos/__main__.py", line 1003, in run_sequential
    test_result = run(setup_ex, run_args.abi, fun_info, test_config)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/src/halmos/__main__.py", line 727, in run
    for idx, ex in enumerate(exs):
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 2651, in run
    self.call(ex, opcode, to_alias, stack, step_id)
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 2064, in call
    call_unknown()
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 1995, in call_unknown
    ret = hevm_cheat_code.handle(self, ex, arg, stack, step_id)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/src/halmos/cheatcodes.py", line 542, in handle
    return ByteVec(sevm.sload(ex, load_account_alias, load_slot))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 1679, in sload
    return self.storage_model.load(ex, addr, loc)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 1297, in load
    cls.init(ex, addr, loc)
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 1290, in init
    assert_address(addr)
  File "/Users/karma/projects/halmos/src/halmos/utils.py", line 478, in assert_address
    elif x.size() != 160:
         ^^^^^^
AttributeError: 'NoneType' object has no attribute 'size'
Symbolic test result: 0 passed; 1 failed; time: 0.03s

[time] total: 1.28s (build: 0.71s, load: 0.53s, tests: 0.04s)
```

`failed()` does the following:

```solidity
    function failed() public returns (bool) {
        if (_failed) {
            return _failed;
        } else {
            bool globalFailed = false;
            if (hasHEVMContext()) {  // <- true since we return extcodesize 1 for the hevm address
                (, bytes memory retdata) = HEVM_ADDRESS.call(
                    abi.encodePacked(
                        bytes4(keccak256("load(address,bytes32)")),
                        abi.encode(HEVM_ADDRESS, bytes32("failed"))
                    )
                );
                globalFailed = abi.decode(retdata, (bool));  // <- fails during the handling of vm.load(hevm, 'failed')
            }
            return globalFailed;
        }
    }
```

In halmos, vm.load is handled like this:

```python
        # vm.load(address,bytes32)
        elif funsig == hevm_cheat_code.load_sig:
            load_account = uint160(arg.get_word(4))  #  == 0x7109709ecfa91a80626ff3989d68f67f5b1dd12d
            load_slot = uint256(arg.get_word(36))
            load_account_alias = sevm.resolve_address_alias(
                ex, load_account, stack, step_id, branching=False
            )

            return ByteVec(sevm.sload(ex, load_account_alias, load_slot))
```

resolve_address_alias returns `None` for `load_account=0x7109709ecfa91a80626ff3989d68f67f5b1dd12d`, which leads to the downstream exception in `sevm.sload`

daejunpark:
fix #358

### DIFF
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -354,6 +354,18 @@ class hevm_cheat_code:
         )
     )
 
+    # abi.encodePacked(
+    #     bytes4(keccak256("load(address,bytes32)")),
+    #     abi.encode(HEVM_ADDRESS, bytes32("failed"))
+    # )
+    failed_payload = ByteVec(
+        bytes.fromhex(
+            "667f9d70"
+            + "0000000000000000000000007109709ecfa91a80626ff3989d68f67f5b1dd12d"
+            + "6661696c65640000000000000000000000000000000000000000000000000000"
+        )
+    )
+
     # bytes4(keccak256("assume(bool)"))
     assume_sig: int = 0x4C63E562
 
@@ -533,6 +545,9 @@ class hevm_cheat_code:
 
         # vm.load(address,bytes32)
         elif funsig == hevm_cheat_code.load_sig:
+            if arg == hevm_cheat_code.failed_payload:
+                return ByteVec(con(0))
+
             load_account = uint160(arg.get_word(4))
             load_slot = uint256(arg.get_word(36))
             load_account_alias = sevm.resolve_address_alias(
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -545,7 +545,10 @@ class hevm_cheat_code:
 
         # vm.load(address,bytes32)
         elif funsig == hevm_cheat_code.load_sig:
+            # vm.load(HEVM_ADDRESS, "failed") is handled separately
             if arg == hevm_cheat_code.failed_payload:
+                # since fail(), which triggers vm.store(HEVM_ADDRESS, "failed", 1), halts immediately, (see vm.store() above)
+                # the "failed" slot is never assigned, thus vm.load(HEVM_ADDRESS, "failed") always return zero at this point
                 return ByteVec(con(0))
 
             load_account = uint160(arg.get_word(4))
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -208,6 +208,11 @@ def symbolic_storage(ex, arg, sevm, stack, step_id):
     account_alias = sevm.resolve_address_alias(
         ex, account, stack, step_id, branching=False
     )
+
+    if account_alias is None:
+        error_msg = f"enableSymbolicStorage() is not allowed for a nonexistent account: {hexify(account)}"
+        raise HalmosException(error_msg)
+
     ex.storage[account_alias].symbolic = True
 
 
@@ -351,18 +356,6 @@ class hevm_cheat_code:
             + "0000000000000000000000007109709ecfa91a80626ff3989d68f67f5b1dd12d"
             + "6661696c65640000000000000000000000000000000000000000000000000000"
             + "0000000000000000000000000000000000000000000000000000000000000001"
-        )
-    )
-
-    # abi.encodePacked(
-    #     bytes4(keccak256("load(address,bytes32)")),
-    #     abi.encode(HEVM_ADDRESS, bytes32("failed"))
-    # )
-    failed_payload = ByteVec(
-        bytes.fromhex(
-            "667f9d70"
-            + "0000000000000000000000007109709ecfa91a80626ff3989d68f67f5b1dd12d"
-            + "6661696c65640000000000000000000000000000000000000000000000000000"
         )
     )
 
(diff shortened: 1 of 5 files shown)


karmacoma-eth:
ah I see it does address the specific issue in https://github.com/a16z/halmos/issues/358, but we still have a more general problem: should `resolve_address_alias` return `Address | None`?

- if it does, it's up to the callers to check if there is a result
- if it doesn't, then we should fix the path where it does return None

src/halmos/cheatcodes.py
```diff
@@ -533,6 +545,12 @@ def handle(sevm, ex, arg: ByteVec, stack, step_id) -> ByteVec | None:
 
         # vm.load(address,bytes32)
         elif funsig == hevm_cheat_code.load_sig:
+            # vm.load(HEVM_ADDRESS, "failed") is handled separately
+            if arg == hevm_cheat_code.failed_payload:
+                # since fail(), which triggers vm.store(HEVM_ADDRESS, "failed", 1), halts immediately, (see vm.store() above)
+                # the "failed" slot is never assigned, thus vm.load(HEVM_ADDRESS, "failed") always return zero at this point
+                return ByteVec(con(0))
```
karmacoma-eth, line 552: would it be bad to return 0 if `load_account_alias` is `None`?

karmacoma-eth: ideally we wouldn't handle hevm as an edge case

daejunpark: good idea! how should we do for vm.store() with nonexistent account?

daejunpark:
> ah I see it does address the specific issue in #358, but we still have a more general problem: should `resolve_address_alias` return `Address | None`?
> 
> * if it does, it's up to the callers to check if there is a result
> * if it doesn't, then we should fix the path where it does return None

yes, it returns None when the given address doesn't exist, or is assumed to not exist. so it's the caller's responsibility to check the result.

daejunpark:
> * if it does, it's up to the callers to check if there is a result

done

src/halmos/cheatcodes.py
daejunpark: done for vm.load().

for vm.store(), currently an exception is raised if the given account doesn't exist.

karmacoma-eth: > for vm.store(), currently an exception is raised if the given account doesn't exist

ok, so the idea is that we need to first deploy code at an address and then we can vm.store to it? is that because deploying resets the storage?

karmacoma-eth (approved):

karmacoma-eth (approved):
very nice, thank you!

src/halmos/cheatcodes.py
daejunpark: so far, we have the invariant that for any address x, x in ex.code iff x in ex.storage.  now, if we allow vm.store() for nonexistent addresses, the invariant would no longer hold, which leads to some logic changes in the codebase.

another invariant we have is that all addresses in ex.code/ex.storage are concrete.  if the given address is symbolic, this invariant would be broken.  note that calls to symbolic addresses don't break this, because of branching over concrete aliases, and calls to nonexistent address are essentially no-op. 

while these invariants are convenient for various reasons, they aren't absolutely necessary. so we can break them and adjust our logic, if needed, although this would increase code complexity.

i'd suggest we wait until we see concrete use cases where vm.store() is used with nonexistent or symbolic addresses, before deciding how to redesign this logic.

PR #360, opened 2024-09-02
feat: support for new cheatcode of generating calldata

daejunpark:
done:
- createCalldata(string contractName)
- createCalldata(string contractName, bool includeViewFunctions)
- createCalldata(string filename, string contractName)
- createCalldata(string filename, string contractName, bool includeViewFunctions)

note:
- `contractName` can be either a contract or an interface name.
- `includeViewFunctions` defaults to false, if not provided.

future work:
- createCalldata(address contract)


src/halmos/__main__.py
```diff
@@ -1569,6 +1504,9 @@ def on_signal(signum, frame):
 
         contract_timer = NamedTimer("time")
 
+        # TODO: fix potential race conditions in the test-parallel mode
+        BuildOutMap().set_build_out_map(build_out_map)
+
```
daejunpark, line 1509: nice to see all tests in the test-parallel mode failed as expected :sweat_smile:

daejunpark: i was wrong. the previous test failure was due to the singleton class being not shared across multi processes, not because of race conditions.

for the same reason, the trace pretty-printing feature doesn't work in the test-parallel mode.

src/halmos/__main__.py
```diff
@@ -958,6 +908,7 @@ def run_parallel(run_args: RunArgs) -> list[TestResult]:
             setup_config,
             with_devdoc(args, fun_info.sig, run_args.contract_json),
             libs,
+            run_args.build_out_map,
```
daejunpark, line 911: the serialization of this additional item seems to be causing the [performance overhead](https://github.com/a16z/halmos/actions/runs/10662927742) (6m vs 3m) in the test-parallel mode.

src/halmos/__main__.py
daejunpark: this particular issue has been fixed by having each process to have its own BuildOutMap, but it causes performance overhead in the test-parallel mode: https://github.com/a16z/halmos/pull/360#discussion_r1740491771

src/halmos/__main__.py
daejunpark: to address the performance issue, we can consider removing unnecessary fields from the build json output before serialization.

src/halmos/cheatcodes.py
```diff
@@ -216,6 +219,58 @@ def symbolic_storage(ex, arg, sevm, stack, step_id):
...
+    for funsig in methodIdentifiers:
+        funname = funsig.split("(")[0]
+        funselector = methodIdentifiers[funsig]
+        funinfo = FunctionInfo(funname, funsig, funselector)
+
+        dyn_param_size = []
+
+        calldata = ByteVec()
+        calldata.append(int(funselector, 16).to_bytes(4, "big"))
+        mk_calldata(abi, funinfo, calldata, dyn_param_size, sevm.options)
```
daejunpark, line 263: todo: there is an issue with calling mk_calldata multiple times, where the generated symbol names are not unique across different mk_calldata calls.

daejunpark: fixed in https://github.com/a16z/halmos/pull/360/commits/ca1abbfba0e3f7643d62dc0d4a5e86e1fe1ddc81

karmacoma-eth:
suggested test:

```python
import pytest
from z3 import BitVec, BitVecSort
from halmos.cheatcodes import permutate_dyn_size, DynamicParams, DynamicArrayType
from halmos.bytevec import ByteVec

# Mock classes and functions
class MockSEVM:
    class Options:
        pass
    options = Options()

class MockExecutor:
    def new_symbol_id(self):
        return 0

class MockFunctionInfo:
    def __init__(self):
        self.name = "testFunction"
        self.signature = "testFunction(uint256[],string)"
        self.selector = "0x12345678"

def test_permutate_dyn_size():
    # Setup
    dyn_param_size = DynamicParams()
    dyn_param_size.append(("param1", 2, DynamicArrayType("uint256")))
    dyn_param_size.append(("param2", 32, str))  # Assuming str represents a dynamic string

    funselector = "0x12345678"
    abi = [
        {
            "name": "testFunction",
            "type": "function",
            "inputs": [
                {"name": "param1", "type": "uint256[]"},
                {"name": "param2", "type": "string"}
            ]
        }
    ]
    funinfo = MockFunctionInfo()
    sevm = MockSEVM()
    ex = MockExecutor()

    # Call the function
    result = permutate_dyn_size(dyn_param_size, funselector, abi, funinfo, sevm, ex)

    # Assertions
    assert len(result) == 12  # 3 options for param1 * 4 options for param2

    # Check that all expected sizes are present
    expected_sizes = {
        "param1": [0, 1, 2],
        "param2": [0, 32, 65, 1024]
    }

    for calldata in result:
        param1_size = int.from_bytes(calldata[36:68], "big")
        param2_size = int.from_bytes(calldata[100:132], "big")
        
        assert param1_size in expected_sizes["param1"]
        assert param2_size in expected_sizes["param2"]

    # Check that all combinations are present
    combinations = set((int.from_bytes(c[36:68], "big"), int.from_bytes(c[100:132], "big")) for c in result)
    assert len(combinations) == 12

    # Check the structure of the calldata
    for calldata in result:
        assert calldata[:4] == b'\x12\x34\x56\x78'  # function selector
        assert len(calldata) >= 132  # minimum length for two dynamic parameters

if __name__ == "__main__":
    pytest.main([__file__])
```

and in `permutate_dyn_size`:

```
        # consider all size combinations
        arrlen_lst = [
            {**arrlen, p_name: new_size}
            for arrlen in arrlen_lst
            for new_size in new_size_options
        ]
```

src/halmos/__main__.py
daejunpark: we discussed and decided to deprecate the test-parallel mode once the portfolio solving feature is ready.

daejunpark:
the cursor-generated test is added with slight modifications: https://github.com/a16z/halmos/pull/360/commits/7c00d34018740f3a0ee19945c31c6eb5c78cd670

daejunpark:
@karmacoma-eth this is now ready to review. note that an interface name turns out to be already supported.

src/halmos/cheatcodes.py
```diff
@@ -216,6 +219,108 @@ def symbolic_storage(ex, arg, sevm, stack, step_id):
...
+            DynamicParams(),
+            sevm.options,
+            arrlen=arrlen,
+            new_symbol_id=ex.new_symbol_id,
+        )
+        result.append(calldata)
+
+    return result
+
+
```
daejunpark, line 323: note: the combination generation logic has an issue with duplication, as shown in HalmosCheatCode test. this will be addressed in a separate pr.

src/halmos/__main__.py
karmacoma-eth: offline discussion: let's get rid of test-parallel mode in future work

examples/simple/test/SimpleState.t.sol
```diff
@@ -0,0 +1,48 @@
...
+    function check_buggy() public {
+        bool success;
+
+        // note: a total of 253 feasible paths are generated, of which only 10 unique states exist
+        for (uint i = 0; i < 10; i++) {
+            (success,) = target.call(svm.createCalldata("SimpleState"));
+            vm.assume(success);
+        }
+
+        assertFalse(buggy());
```
karmacoma-eth, line 46: nit: make counter public, and replace assertion with this:

```suggestion
        assertNotEq(target.counter(), 10);
```

examples/tokens/ERC721/test/OpenZeppelinERC721.t.sol
```diff
@@ -41,4 +41,9 @@ contract OpenZeppelinERC721Test is ERC721Test {
         vm.prank(accounts[1]);
         token_.setApprovalForAll(accounts[2], true);
     }
+
+    function check_NoBackdoor() public {
+        bytes memory _calldata = svm.createCalldata("OpenZeppelinERC721");
```
karmacoma-eth, line 46: very nice 👌

examples/simple/test/SimpleState.t.sol
daejunpark: i agree. it will be done after merging the feature of excluding view functions in createCalldata.

src/halmos/calldata.py
```diff
@@ -79,33 +80,61 @@ def parse_tuple_type(var: str, items: list[dict]) -> Type:
...
 class EncodingResult:
     data: list[BitVecRef]
-    size: int
-    static: bool
+    size: int  # number of bytes
+    static: bool  # static vs dynamic type
+
+
+@dataclass(frozen=True)
+class DynamicParams:
```
karmacoma-eth, line 88: would we be better off with a `DynamicParam` dataclass?

```
@dataclass(frozen=True)
class DynamicParam:
    name: str
    size: int
    typ: Type
```

and `DynamicParams` becomes just a `list[DynamicParam]`?

src/halmos/calldata.py
```diff
@@ -150,27 +179,35 @@ def encode(self, name: str, typ: Type) -> EncodingResult:
...
                 [con(array_len)] + encoded.data, 32 + encoded.size, False
             )
 
         if isinstance(typ, BaseType):
+
+            def new_symbol() -> str:
+                if self.new_symbol_id is None:
+                    return f"p_{name}_{typ.typ}"
+                else:
+                    return f"p_{name}_{typ.typ}_{self.new_symbol_id():>02}"
```
karmacoma-eth, line 195:
```suggestion
            def new_symbol() -> str:
                id_suffix = f"_{self.new_symbol_id()}" if self.new_symbol_id else ""
                return f"p_{name}_{typ.typ}{id_suffix}"
```

src/halmos/calldata.py
```diff
@@ -210,3 +247,71 @@ def head_size(x):
...
+
+
+@dataclass(frozen=True)
+class FunctionInfo:
+    name: str | None = None
+    sig: str | None = None
+    selector: str | None = None
+
+
+def str_abi(item: dict) -> str:
```
karmacoma-eth, line 259: not super clear what the contract for this is given the type info and lack of comments

src/halmos/calldata.py
```diff
@@ -210,3 +247,71 @@ def head_size(x):
...
+    args: HalmosConfig,
+    arrlen: dict[str, int] = None,
+    new_symbol_id: Callable = None,
+) -> None:
+    # find function abi
+    fun_abi = find_abi(abi, fun_info)
+
+    # no parameters
+    if len(fun_abi["inputs"]) == 0:
+        return
```
karmacoma-eth, line 302:
```suggestion
    if not fun_abi["inputs"]:
        return
```

src/halmos/calldata.py
```diff
@@ -210,3 +247,71 @@ def head_size(x):
...
+
+def mk_calldata(
+    abi: list,
+    fun_info: FunctionInfo,
+    cd: ByteVec,
+    dyn_param_size: DynamicParams,
+    args: HalmosConfig,
+    arrlen: dict[str, int] = None,
+    new_symbol_id: Callable = None,
+) -> None:
```
karmacoma-eth, line 296: would it be possible to have `mk_calldata` return `cd` instead of mutating an input parameter?

src/halmos/calldata.py
```diff
@@ -210,3 +247,71 @@ def head_size(x):
...
+            else:
+                ret.append(typ)
+        return "(" + ",".join(ret) + ")"
+
+    if item["type"] != "function":
+        raise ValueError(item)
+    return item["name"] + str_tuple(item["inputs"])
+
+
+def find_abi(abi: list, fun_info: FunctionInfo) -> dict:
```
karmacoma-eth, line 276: I haven't looked precisely, but this seems inefficient, no? We go from a function signature and iterate over abi items until we find one that matches? Is that because that's how the solc json output is stored?

Should we consider building our own index of funsig to funabi?

src/halmos/calldata.py
```diff
@@ -210,3 +247,71 @@ def head_size(x):
...
+
+
+def mk_arrlen(args: HalmosConfig) -> dict[str, int]:
+    arrlen = {}
+    if args.array_lengths:
+        for assign in [x.split("=") for x in args.array_lengths.split(",")]:
+            name = assign[0].strip()
+            size = assign[1].strip()
+            arrlen[name] = int(size)
+    return arrlen
```
karmacoma-eth, line 317: dict comprehension would nice here IMO:

```suggestiondef mk_arrlen(args: HalmosConfig) -> dict[str, int]:
    if not args.array_lengths:
        return {}

    name_size_pairs = args.array_lengths.split(",")
    return {
        name.strip(): int(size.strip())
        for name, size in (x.split("=") for x in name_size_pairs)
    }
```

daejunpark: accepted you suggestion in https://github.com/a16z/halmos/pull/360/commits/7124320edcaf4534395ddec2dddf2f8c70f1b236, but replaced the tuple-generator comprehension with list comprehension. please let me know if the tuple generator was intentional and performs better than list comprehension here.

src/halmos/calldata.py
daejunpark: added comment

src/halmos/calldata.py
daejunpark: agree. yes, abi items are given as list in json output.

btw, this is not new, but existing code moved here. improving it can be done in a separate pr. added todo comment for that.

src/halmos/cheatcodes.py
daejunpark: fixed in https://github.com/a16z/halmos/pull/360/commits/36ecb35bdd86e86093ceb08912c9087c8e3dd365

src/halmos/calldata.py
daejunpark: done in https://github.com/a16z/halmos/pull/360/commits/36ecb35bdd86e86093ceb08912c9087c8e3dd365

src/halmos/calldata.py
daejunpark: will be done in a separate pr. added a todo comment for that.

daejunpark:
@karmacoma-eth this is ready for re-review. the multiple calldata encoding for dynamic sizes has been re-implemented to address issues in the previous approach.

karmacoma-eth (approved):

src/halmos/mapper.py
```diff
@@ -76,6 +79,55 @@ def __call__(cls, *args, **kwargs):
...
+        """
+        if not self._build_out_map_reverse:
+            self.create_build_out_map_reverse()
+
+        mapping = self._build_out_map_reverse[contract_name]
+
+        if not mapping:
+            raise HalmosException(f"{contract_name} is not found")
+
+        if filename is None:
```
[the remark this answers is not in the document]
daejunpark: filename could be empty string

tests/regression/test/HalmosCheatCode.t.sol
```diff
@@ -138,6 +138,148 @@ contract HalmosCheatCodeTest is SymTest, Test {
         // symbolic storage is not allowed for a nonexistent account
         svm.enableSymbolicStorage(address(0xdeadbeef)); // HalmosException
     }
+
+    function check_createCalldata_Beep_1_excluding_pure() public {
+        bytes memory data = svm.createCalldata("HalmosCheatCode.t.sol", "Beep");
+        _check_createCalldata_Beep(data); // fail // not reachable, no calldata generated above
+    }
```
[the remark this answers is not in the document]
daejunpark: it is expected to fail because the only function is pure and it is excluded in the calldata generation.

agree that the error message needs to be improved. it should say something like "no success (or normally terminated) path exists."

tests/regression/test/HalmosCheatCode.t.sol
```diff
@@ -138,6 +138,97 @@ contract HalmosCheatCodeTest is SymTest, Test {
...
+        console.log("zoo");
+        // 12 (= 3 * 4) combinations
+        console.log(x.length); // 0, 1, 2
+        console.log(y.length); // 0, 32, 65, 1024
+        return this.zoo.selector;
+    }
+
+    function foobar(bytes[] memory x) public returns (bytes4) {
+        console.log("foobar");
+        // TODO: currently it returns 48 (= 16*3) calldata combinations, while it should do only 21 (= 16+4+1) combinations.
```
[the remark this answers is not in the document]
daejunpark: 4*4. when the size of x is 2, there are two bytes elements. for each byte elements, we have four size candidates, 0, 32, 65, 1024.

src/halmos/calldata.py
```diff
@@ -79,33 +80,61 @@ def parse_tuple_type(var: str, items: list[dict]) -> Type:
...
+        self,
+        args: HalmosConfig,
+        arrlen: dict[str, int],
+        dyn_param_size: DynamicParams,
+        new_symbol_id: Callable,
     ) -> None:
         self.args = args
         self.arrlen = arrlen
         self.dyn_param_size = dyn_param_size
+        self.new_symbol_id = new_symbol_id
```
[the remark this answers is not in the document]
daejunpark: to generate unique symbol names for multiple create_calldata calls

src/halmos/calldata.py
daejunpark: done

src/halmos/calldata.py
daejunpark: done

src/halmos/calldata.py
```diff
@@ -210,3 +245,72 @@ def head_size(x):
...
+        ):
+            return item
+    raise ValueError(f"No {funsig} found in {abi}")
+
+
+def mk_calldata(
+    abi: list,
+    fun_info: FunctionInfo,
+    cd: ByteVec,
+    dyn_param_size: DynamicParams,
```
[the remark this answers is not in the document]
daejunpark: done in https://github.com/a16z/halmos/pull/360/commits/2256fb3d5c2944a5ffd3dfaccb23aeacb64444fe

src/halmos/calldata.py
```diff
@@ -210,3 +245,72 @@ def head_size(x):
...
+    args: HalmosConfig,
+    arrlen: dict[str, int] = None,
```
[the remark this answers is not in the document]
daejunpark: done in https://github.com/a16z/halmos/pull/360/commits/2256fb3d5c2944a5ffd3dfaccb23aeacb64444fe

src/halmos/sevm.py
```diff
@@ -2066,38 +2067,49 @@ def call_unknown() -> None:
...
-                        data=ret,
-                        error=None,
-                    ),
-                    depth=ex.context.depth + 1,
+            ret_lst = ret if isinstance(ret, list) else [ret]
+
+            for idx, ret_ in enumerate(ret_lst):
+                new_ex = (
+                    self.create_branch(ex, BoolVal(True), ex.pc)
+                    if idx < len(ret_lst) - 1
```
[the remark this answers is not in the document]

src/halmos/cheatcodes.py
```diff
@@ -216,6 +219,108 @@ def symbolic_storage(ex, arg, sevm, stack, step_id):
...
+    )
+
+
+def create_calldata_contract(ex, arg, sevm, stack, step_id):
+    contract_name = name_of(extract_string_argument(arg, 0))
+    return create_calldata_generic(ex, arg, sevm, stack, step_id, contract_name)
+
+
+def create_calldata_generic(
+    ex, arg, sevm, stack, step_id, contract_name, filename=None
```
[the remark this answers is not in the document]
daejunpark: removed

src/halmos/calldata.py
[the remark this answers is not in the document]
daejunpark: added comment in https://github.com/a16z/halmos/pull/360/commits/2256fb3d5c2944a5ffd3dfaccb23aeacb64444fe

src/halmos/calldata.py
```diff
@@ -79,33 +80,61 @@ def parse_tuple_type(var: str, items: list[dict]) -> Type:
...
         self.args = args
         self.arrlen = arrlen
         self.dyn_param_size = dyn_param_size
+        self.new_symbol_id = new_symbol_id
 
-    def choose_array_len(self, name: str) -> int:
+    def choose_array_len(self, name: str, typ: Type) -> int:
         if name in self.arrlen:
             array_len = self.arrlen[name]
         else:
```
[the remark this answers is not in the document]
daejunpark: done

src/halmos/mapper.py
```diff
@@ -76,6 +79,55 @@ def __call__(cls, *args, **kwargs):
...
+            if len(mapping) > 1:
+                err_msg = f"{contract_name} exists in multiple files: {mapping.keys()}"
```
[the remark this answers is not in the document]
daejunpark: done

src/halmos/mapper.py
```diff
@@ -76,6 +79,55 @@ def __call__(cls, *args, **kwargs):
...
+                raise HalmosException(err_msg)
+            filename = list(mapping.keys())[0]
```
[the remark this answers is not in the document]
daejunpark: done

src/halmos/mapper.py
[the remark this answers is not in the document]
karmacoma-eth: what would be the meaning of that? isn't it just going to fail with `"{contract_name} is not found in ''"`?

tests/regression/test/HalmosCheatCode.t.sol
```diff
@@ -138,6 +138,148 @@ contract HalmosCheatCodeTest is SymTest, Test {
         // symbolic storage is not allowed for a nonexistent account
         svm.enableSymbolicStorage(address(0xdeadbeef)); // HalmosException
     }
+
+    function check_createCalldata_Beep_1_excluding_pure() public {
+        bytes memory data = svm.createCalldata("HalmosCheatCode.t.sol", "Beep");
+        _check_createCalldata_Beep(data); // fail // not reachable, no calldata generated above
+    }
+
+    function check_createCalldata_Beep_1() public {
```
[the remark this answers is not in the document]
daejunpark: done in https://github.com/a16z/halmos/pull/360/commits/0b3847d18faa7445fe2ea83ad9c189fe92e2a417

src/halmos/mapper.py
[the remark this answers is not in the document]
daejunpark: yes, that's why this branch shouldn't be taken when filename is empty. otherwise, confusing error messages may be generated.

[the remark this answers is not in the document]
daejunpark: i mean, i want to avoid that this error message `"{contract_name} exists in multiple files: {list(mapping.keys())}"` may be printed when filename is empty.

daejunpark:
@karmacoma-eth i've addressed all your comments, and implemented some todo items. this should be enough for the scope of this pr. please take another look when you get a chance.

karmacoma-eth (approved):
looks great 💪

PR #361, opened 2024-09-14
perf: support for custom action in config parser

daejunpark:
add support for custom action in halmos config. this allows structural option values (e.g., --array-length) to be parsed at the time of config creation.

note: since the config object is frozen, it is nontrivial to extend or update the config with parsing raw string values after config has been created. also, since config objects are re-generated to override further configurations, it's more convenient to parse raw string through argparser.

### DIFF
--- a/src/halmos/config.py
+++ b/src/halmos/config.py
@@ -52,9 +52,9 @@ def arg(
     )
 
 
-class ParseCSVAction(argparse.Action):
+class ParseCSV(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
-        values = ParseCSVAction.parse(values)
+        values = ParseCSV.parse(values)
         setattr(namespace, self.dest, values)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


src/halmos/config.py
```diff
@@ -45,10 +47,39 @@ def arg(
             "short": short,
             "countable": countable,
             "global_default_str": global_default_str,
+            "action": action,
```
daejunpark, line 50: would it be better to use a different name, e.g., `custom_action`, to avoid potential name clashes? it might be fine as the name `choices` is also used here, but let me know if you have any concerns.

karmacoma-eth: not sure I follow, name clash with what? 

I'm ok with `custom_action`, but don't love the "action" nomenclature. If this is about parsing, what about `parser`?

src/halmos/config.py
```diff
@@ -45,10 +47,39 @@ def arg(
...
         },
     )
 
 
+class ParseCSV(argparse.Action):
```
karmacoma-eth, line 55: what do we get by subclassing `argparse.Action`?

src/halmos/config.py
karmacoma-eth: `parser` is even more overloaded, ignore that. Either `action` or `custom_action` works IMO, I wouldn't worry about name clashes at that level (I thought this was a concern about shadowing)

src/halmos/config.py
karmacoma-eth: ah I see we pass it to the generated argparse parser. Pretty cool!

karmacoma-eth (approved):

src/halmos/config.py
daejunpark: okay, i will stick with `action`.

PR #364, opened 2024-09-17
fix: concretize calldatacopy

daejunpark:
there are two instructions that read calldata, and their concretization logic works as follows:
- calldataload: concretizes symbols if they exist in either the substitution or candidates mapping. it may cause the current path to branch.
- calldatacopy: concretizes symbols only if they exist in the substitution mapping. no path branching.

this behavior is based on typical calldata decoding logic, where the length of dynamic arguments is first read using calldataload, and their data is then read using either calldataload or calldatacopy.

### DIFF
--- a/src/halmos/bytevec.py
+++ b/src/halmos/bytevec.py
@@ -5,6 +5,7 @@ from typing import Any, ForwardRef
 
 from sortedcontainers import SortedDict
 from z3 import BitVecRef, If, eq, is_bool, is_bv, is_bv_value, simplify
+from z3.z3util import is_expr_var
 
 from .utils import (
     Byte,
@@ -609,6 +610,23 @@ class ByteVec:
             value = If(value, con(1), con(0))
 
         self.set_slice(offset, offset + 32, value)
+
+    def concretize(self, substitution: dict[BitVecRef, BitVecRef]) -> None:
+        """
+        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substituion mapping.
+        """
+        for offset, chunk in self.chunks.items():
+            chunk_data = chunk.data
+            if not isinstance(chunk, SymbolicChunk) or not is_expr_var(chunk_data):
+                continue
+
+            concrete_chunk_data = substitution.get(chunk_data)
+            if concrete_chunk_data is None:  # could be zero
+                continue
+
+            self.chunks[offset] = ConcreteChunk(
+                bv_value_to_bytes(concrete_chunk_data), chunk.start, chunk.length
+            )
 
     ### read operations
 
--- a/src/halmos/bytevec.py
+++ b/src/halmos/bytevec.py
@@ -613,7 +613,7 @@ class ByteVec:
 
     def concretize(self, substitution: dict[BitVecRef, BitVecRef]) -> None:
         """
-        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substituion mapping.
+        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substitution mapping.
         """
         for offset, chunk in self.chunks.items():
             chunk_data = chunk.data
--- a/src/halmos/bytevec.py
+++ b/src/halmos/bytevec.py
@@ -613,7 +613,10 @@ class ByteVec:
 
     def concretize(self, substitution: dict[BitVecRef, BitVecRef]) -> None:
         """
-        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substitution mapping.
+        Replace all top-level symbols in the chunks with their corresponding concrete values, if they exist in the given substitution mapping.
+
+        Note: only top-level symbols are currently replaced, not those within nested symbolic terms.
+        This approach is sufficient for current calldata use cases. Performance impacts should be evaluated if all symbol occurrences need to be replaced.
         """
         for offset, chunk in self.chunks.items():
             chunk_data = chunk.data
--- a/src/halmos/bytevec.py
+++ b/src/halmos/bytevec.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 from typing import Any, ForwardRef
 
 from sortedcontainers import SortedDict
-from z3 import BitVecRef, If, eq, is_bool, is_bv, is_bv_value, simplify
+from z3 import BitVecRef, If, eq, is_bool, is_bv, is_bv_value, simplify, substitute
 from z3.z3util import is_expr_var
 
 from .utils import (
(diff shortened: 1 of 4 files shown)


src/halmos/bytevec.py
```diff
@@ -610,6 +611,23 @@ def set_word(self, offset: int, value: Word) -> None:
 
         self.set_slice(offset, offset + 32, value)
 
+    def concretize(self, substitution: dict[BitVecRef, BitVecRef]) -> None:
+        """
+        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substituion mapping.
```
karmacoma-eth, line 616: typo: 

```suggestion
        Replace all symbols in the chunks with their corresponding concrete values, if they exist in the given substitution mapping.
```

src/halmos/bytevec.py
```diff
@@ -610,6 +611,23 @@ def set_word(self, offset: int, value: Word) -> None:
...
+        """
+        for offset, chunk in self.chunks.items():
+            chunk_data = chunk.data
+            if not isinstance(chunk, SymbolicChunk) or not is_expr_var(chunk_data):
```
karmacoma-eth, line 620: what is the purpose of `is_expr_var` here?

src/halmos/sevm.py
```diff
@@ -644,8 +644,10 @@ class Concretization:
     In the future, it may be used for other purposes, such as concrete hash reasoning.
     """
 
-    substitution: dict = field(default_factory=dict)  # symbol -> constant
-    candidates: dict = field(default_factory=dict)  # symbol -> set of constants
+    # symbol -> constant
+    substitution: dict[BitVecRef, BitVecRef] = field(default_factory=dict)
```
karmacoma-eth, line 648: is this actually `substitution: dict[BitVecRef, int]`?

src/halmos/bytevec.py
```diff
@@ -610,6 +611,23 @@ def set_word(self, offset: int, value: Word) -> None:
...
+                continue
+
+            concrete_chunk_data = substitution.get(chunk_data)
+            if concrete_chunk_data is None:  # could be zero
+                continue
+
+            self.chunks[offset] = ConcreteChunk(
+                bv_value_to_bytes(concrete_chunk_data), chunk.start, chunk.length
+            )
```
karmacoma-eth, line 629: caveat: this may not work if the symbol is fragmented over 2 or more consecutive chunks, right?

src/halmos/sevm.py
karmacoma-eth: are there any constraints on the bit size of keys in `substitution`?

karmacoma-eth (approved):

src/halmos/bytevec.py
daejunpark: it checks if it's simply a symbol, rather than an arbitrary symbolic term.

this is enough for the current calldata use cases. later, we may consider concretizing all symbols inside terms, but we'll need to make sure it doesn't hurt performance.

karmacoma-eth: cool, do you want to add it as a comment? I will definitely forget this :)

src/halmos/sevm.py
daejunpark: no, the values are bv_value. no constraints on the bitsize.

src/halmos/bytevec.py
daejunpark: i think it should work. why do you think it may not?

daejunpark: each chunk preserves its offset and length, and then they will be defraged later during wrap(), right?

src/halmos/bytevec.py
daejunpark: added: https://github.com/a16z/halmos/pull/364/commits/9c46197d9263f24676ea0231961ec821b3f9f0a2

src/halmos/bytevec.py
karmacoma-eth: here is what I mean

```py
x = BitVec('x', 256)
substitutions = { x : uint256(42) }

# for some reason, x has been split over 2 chunks (maybe 2 separate MSTOREs)
data = ByteVec()
data.append(Extract(256, 128, x))
data.append(Extract(128, 0, x))

data.concretize(substitutions)  # x won't be substituted

# this may seem weird because this is true:
eq(data.unwrap(), x)
```

src/halmos/bytevec.py
```diff
@@ -610,6 +611,26 @@ def set_word(self, offset: int, value: Word) -> None:
...
+        for offset, chunk in self.chunks.items():
+            chunk_data = chunk.data
+            if not isinstance(chunk, SymbolicChunk) or not is_expr_var(chunk_data):
+                continue
+
+            concrete_chunk_data = substitution.get(chunk_data)
+            if concrete_chunk_data is None:  # could be zero
+                continue
+
+            self.chunks[offset] = ConcreteChunk(
```
karmacoma-eth, line 630: also are you sure we want to mutate in place vs return a concretized (but different) ByteVec object?

src/halmos/bytevec.py
karmacoma-eth: but I imagine `is_expr_var` would filter the `Extracts` anyway

src/halmos/bytevec.py
daejunpark: i actually considered returning a new object, but worried about performance. it's safer to return a new object though, to be used in different use cases. wdyt?

src/halmos/bytevec.py
daejunpark: ah, if x is substituted inside extract, unwrap() should return 42, right?

daejunpark: fixed in https://github.com/a16z/halmos/pull/364/commits/c18fbc586af2cd024425736df2d22eca9cdd2276

src/halmos/bytevec.py
daejunpark: fixed to return a new object: https://github.com/a16z/halmos/pull/364/commits/c18fbc586af2cd024425736df2d22eca9cdd2276

src/halmos/config.py
```diff
@@ -549,7 +549,15 @@ def parse_dict(self, parsed: dict, source: str = "halmos.toml") -> dict:
...
-        return {k.replace("-", "_"): v for k, v in data.items()}
+        def parse_value(k, v):
+            if k == "default-bytes-lengths":
+                return ParseCSV.parse(v)
+            elif k == "array-lengths":
+                return ParseArrayLengths.parse(v)
+            else:
+                return v
+
+        return {k.replace("-", "_"): parse_value(k, v) for k, v in data.items()}
```
daejunpark, line 560: @karmacoma-eth this logic requires to be updated whenever a new custom parser is added. is there an efficient way to search the custom parser, possibly avoiding a linear scan of `fields(Config)`?

src/halmos/bytevec.py
```diff
@@ -142,6 +142,33 @@ def __getitem__(self, key) -> Byte | ForwardRef("Chunk"):
...
+
+        Return a new object with concrete values, or self if no substitution is made.
+        """
+
+        if not isinstance(self, SymbolicChunk):
+            return self
+
+        data = self.data
+        # TODO: call substitute() only when the data contains variables to be substituted
+        new_data = substitute(data, *substitution.items())
```
karmacoma-eth, line 158: I am slightly worried about the perf impact of `substitute` always constructing a new object, but I guess we'll see how it looks like when I profile (not a blocker)

karmacoma-eth (approved):

src/halmos/bytevec.py
daejunpark: yeah, proper profiling would be greatly helpful. based on my performance tests, it gets slower if data is a complex term, e.g., LibUint1024Test:testProveSubAdd().

i'm thinking applying this substitution to selected `int_of()` only, rather than blindly applying to all calldatacopy. would be a separate pr though.

src/halmos/config.py
```diff
@@ -549,7 +549,17 @@ def parse_dict(self, parsed: dict, source: str = "halmos.toml") -> dict:
                 )
                 sys.exit(2)
 
-        return {k.replace("-", "_"): v for k, v in data.items()}
+        # gather custom actions
+        actions = {
+            field.name: field.metadata.get("action", None) for field in fields(Config)
```
karmacoma-eth, line 554: instead of having a mostly `None` dict, could filter entries:

```suggestion
            field.name: field.metadata["action"] for field in fields(Config) if "action" in field.metadata
```

that way your actions dict contain only the fields with a custom action

karmacoma-eth (approved):

PR #366, opened 2024-09-19
fix: manually solve dynamic array overflow conditions

daejunpark:
this pr fixes the false positive warnings of loop unrolling bounds, which were generated during the creation of storage arrays:
```
contract ATest {
    uint[] numbers;

    function setUp() {
        numbers = new uint[](5);
        ...
    }
}
```


src/halmos/sevm.py
```diff
@@ -994,15 +996,35 @@ def dump(self, print_mem=False) -> str:
     def advance_pc(self) -> None:
         self.pc = self.pgm.next_pc(self.pc)
 
-    def check(self, cond: Any) -> Any:
-        cond = simplify(cond)
+    def quick_custom_check(self, cond: BitVecRef) -> CheckSatResult | None:
```
karmacoma-eth, line 999: nice :ok

src/halmos/utils.py
```diff
@@ -349,6 +353,40 @@ def byte_length(x: Any, strict=True) -> int:
...
+
+    This condition is satisfied when a dynamic array at `slot` exceeds the storage limit.
+    Since such an overflow is highly unlikely in practice, we assume that this condition is unsat.
+
+    Note: we already assume that any sha3 hash output is smaller than 2**256 - 2**64 (see SEVM.sha3_data()).
+    However, the smt solver may not be able to solve this condition within the branching timeout.
+    In such cases, this explicit pattern serves as a fallback to avoid exploring practically infeasible paths.
+
+    We don't need to handle the negation of this condition, because unknown conditions are conservatively assumed to be sat.
+    """
```
karmacoma-eth, line 369: appreciate the detailed reasoning ♥️

src/halmos/utils.py
```diff
@@ -349,6 +353,40 @@ def byte_length(x: Any, strict=True) -> int:
...
+        return False
+    ule = cond.arg(0)
+
+    # Not(ULE(left, right)
+    if not is_app_of(ule, Z3_OP_ULEQ):
+        return False
+    left, right = ule.arg(0), ule.arg(1)
+
+    # Not(ULE(f_sha3_256(slot), offset + base))
+    if not (left.decl().name() == "f_sha3_256" and is_app_of(right, Z3_OP_BADD)):
```
karmacoma-eth, line 382: nit: this might inadvertently break if the `"f_sha3_256"` name changes, can we use some well defined symbol rather than a magic string?

karmacoma-eth (approved):

src/halmos/utils.py
daejunpark: fixed in https://github.com/a16z/halmos/pull/366/commits/edd2dea7d29732002f26bdccbdd12ffa35e6b489

PR #368, opened 2024-09-21
perf: increase default assertion solving timeout from 1s to 1m

daejunpark:
the current default timeout for assertion solving is 1s, which is too short for most cases. as a result, many users disable the timeout altogether. this pr updates the default timeout to a more reasonable value, allowing most tests to be solved within the default time limit.


karmacoma-eth (approved):
makes sense, I like it. Only concern is that 1m maybe feels a little high, my vote would be for something like 10s-20s (enough time to give result, but not long enough that it feels stuck)

daejunpark:
fyi, the 1m limit was chosen based on the solving time needed for those tests in our ci, on M1 pro.

i agree with your concern, but i think even 10s might feel getting stuck especially when there are many paths, unless proper communication of progress is provided. a better approach to mitigate would be to provide clearer progress updates.

PR #378, opened 2024-09-26
feat: added uv install method to readme

PatrickAlphaC:


karmacoma-eth:
thanks @PatrickAlphaC! we were just talking about this yesterday and we were saying that we probably want the default installation instruction to be with `uv` indeed 😁

karmacoma-eth (approved):

PR #382, opened 2024-10-01
feat: target Panic(k) for any k specified in --panic-error-codes

daejunpark:
Previously, only Panic(1) was treated as a test failure that triggered counterexample generation, meaning other types of reverts were silently ignored. This made sense because we were focused on detecting assertion failures in test contracts. However, there are cases where we'd like to ensure that not only assertion failures but also other types of reverts (e.g., arithmetic overflows) do not occur.

Now, Panic(k) for any arbitrary k specified using the new `--panic-error-codes` flag is treated as a test failure. The default value for this flag is 1, maintaining the previous default behavior.

The flag takes a list of error codes (in either decimal or hex format.) To match any integer k, you can use `--panic-error-codes *`.

--

For example, the following test will fail with a counterexample due to overflow:
```
    /// @custom:halmos --panic-error-codes 0x11
    function check_panic_0x11_fail(uint x) public returns (uint) {
        // 0x11: overflow
        return x + 1; // counterexample: x == 2^256 - 1
    }
```
See [`tests/regression/test/Panic.t.sol`](https://github.com/a16z/halmos/blob/main/tests/regression/test/Panic.t.sol) for more examples.

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -132,7 +132,7 @@ ZERO, ONE = con(0), con(1)
 MAX_CALL_DEPTH = 1024
 
 # bytes4(keccak256("Panic(uint256)"))
-PANIC_SELECTOR = ByteVec(bytes.fromhex("4E487B71"))
+PANIC_SELECTOR = bytes.fromhex("4E487B71")
 
 EMPTY_BALANCE = Array("balance_00", BitVecSort160, BitVecSort256)
 
@@ -924,7 +924,7 @@ class Exec:  # an execution path
         if byte_length(error_data) != 36:
             return False
 
-        error_selector = error_data[0:4]
+        error_selector = error_data[0:4].unwrap()
         if error_selector != PANIC_SELECTOR:
             return False
 
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -906,7 +906,7 @@ class Exec:  # an execution path
     def is_halted(self) -> bool:
         return self.context.output.data is not None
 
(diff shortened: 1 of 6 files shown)


src/halmos/sevm.py
```diff
@@ -903,14 +906,36 @@ def halt(
...
+
+        if not isinstance(output.error, Revert):
+            return False
+
+        error_data = output.data
+        if byte_length(error_data) != 36:
+            return False
+
+        error_selector = error_data[0:4]
+        if error_selector != PANIC_SELECTOR:
```
karmacoma-eth, line 928: slightly more efficient to do this:

```py
PANIC_SELECTOR = bytes.fromhex("4E487B71")
...
if PANIC_SELECTOR != error_selector.unwrap()
```

this would avoid repeatedly unwrapping PANIC_SELECTOR and quickly return false if error_selector is not `bytes`

karmacoma-eth (approved):
very nice, I love this! might just want to give `tests/regression/halmos.toml` a little update before merging

src/halmos/sevm.py
daejunpark: nice. forgot that bytevec comparison involves unwrap(). fixed in https://github.com/a16z/halmos/pull/382/commits/825770d7e1adf7ef5b443d9cb23c9872189af7ae

daejunpark:
halmos.toml has been updated: https://github.com/a16z/halmos/pull/382/commits/aec0f9ee5b06567887efda7315060d9b7629c579

PR #383, opened 2024-10-01
fix: unparse config value to generate toml file

daejunpark:
Certain option values (e.g., `--panic-error-codes`) are of non-primitive types, and not all of these types can be represented in TOML syntax, e.g., sets.

This PR ensures that these non-primitive type values are unparsed to their string representation when generating a TOML config file, so that they can be parsed back later when processing the config file.

### DIFF
--- a/src/halmos/config.py
+++ b/src/halmos/config.py
@@ -59,7 +59,7 @@ def ensure_non_empty(values: list | set | dict, raw_values: str) -> list:
 
 
 def parse_csv(values: str, sep: str = ",") -> Generator[Any, None, None]:
-    """Parse a CSV string and return a generator of values."""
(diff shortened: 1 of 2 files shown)


karmacoma-eth (approved):
makes sense, LGTM

src/halmos/config.py
```diff
@@ -75,7 +90,13 @@ def parse(values: str) -> set[int]:
             return set()
 
         # support multiple bases: decimal, hex, etc.
-        return set(int(x.strip(), 0) for x in values.split(","))
+        return ensure_non_empty(set(int(x, 0) for x in parse_csv(values)), values)
```
karmacoma-eth, line 93: TIL:

> Base 0 means to interpret the base from the string as an integer literal.

Issue #283, opened 2024-05-04
support multiple values for --array-lengths option


daejunpark:
support multiple array lengths at once, e.g., (syntax to be decided), `halmos --array-lengths x={1,2,3}`

this should be equivalent to running halmos multiple times for each length value:
```
halmos --array-lengths x=1
halmos --array-lengths x=2
halmos --array-lengths x=3
```

daejunpark:
fixed by #384

PR #384, opened 2024-10-02
feat: consider various lengths for dynamic-sized parameters

daejunpark:
Previously, only a single length for dynamic-sized parameters in test functions was considered.

Now, users can specify multiple lengths, and all possible combinations of these lengths will be considered during symbolic execution. Also, users can now set default lengths for dynamic arrays independently of the loop unrolling bound, whereas they were previously tied to the --loop value implicitly.

Flags to specify lengths:
- --array-lengths: specifies lengths for dynamic parameters (including dynamic arrays, bytes, and strings)
  - syntax: `name1={size1,size2,...},name2=...,` #283
- --default-array-lengths: set default lengths for dynamic arrays (default: `0,1,2`)
- --default-bytes-lengths: set default lengths for bytes and strings (default: `0,65,1024`)

### DIFF
--- a/src/halmos/config.py
+++ b/src/halmos/config.py
@@ -118,10 +118,10 @@ class ParseArrayLengths(argparse.Action):
 
         # check if the format is correct
         # note that the findall pattern below is not sufficient for this check
-        if not re.match(r"^(\w+=(\{[\d,]+\}|\d+)(,|$))*$", values):
+        if not re.match(r"^([^=,\{\}]+=(\{[\d,]+\}|\d+)(,|$))*$", values):
             raise ValueError(f"invalid array lengths format: {values}")
 
-        matches = re.findall(r"(\w+)=(?:\{([\d,]+)\}|(\d+))", values)
+        matches = re.findall(r"([^=,\{\}]+)=(?:\{([\d,]+)\}|(\d+))", values)
         return {
             name.strip(): ensure_non_empty(
                 [int(x) for x in parse_csv(sizes_lst or single_size)]
--- a/src/halmos/config.py
+++ b/src/halmos/config.py
@@ -243,7 +243,7 @@ class Config:
     )
 
     array_lengths: str = arg(
-        help="specify lengths for dynamic-sized arrays, bytes, and string. Lengths can be specified as a comma-separated list of integers enclosed in curly braces, or as a single integer.",
+        help="specify lengths for dynamic-sized arrays, bytes, and string types. Lengths can be specified as a comma-separated list of integers enclosed in curly braces, or as a single integer.",
         global_default=None,
         metavar="NAME1={LENGTH1,LENGTH2,...},NAME2=LENGTH3,...",
         action=ParseArrayLengths,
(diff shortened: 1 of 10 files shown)


src/halmos/config.py
```diff
@@ -109,18 +110,29 @@ def parse(values: str | None) -> dict[str, list[int]]:
         if not values:
             return {}
 
-        # TODO: update syntax: name1={size1,size2},name2=size3,...
+        # syntax: --array-lengths name1=sizes1,name2=sizes2,...
+        # where sizes is either a comma-separated list of integers enclosed in curly braces, or a single integer
+
+        # remove all white spaces to simplify the pattern matching
+        values = "".join(values.split())
```
karmacoma-eth, line 117: I don't understand this line, we don't expect whitespace in the middle of the values, right? why not just `strip()`?

src/halmos/config.py
```diff
@@ -231,15 +243,22 @@ class Config:
     )
 
     array_lengths: str = arg(
-        help="set the length of dynamic-sized arrays including bytes and string (default: loop unrolling bound)",
+        help="specify lengths for dynamic-sized arrays, bytes, and string. Lengths can be specified as a comma-separated list of integers enclosed in curly braces, or as a single integer.",
```
karmacoma-eth, line 246:
```suggestion
        help="specify lengths for dynamic-sized arrays, bytes, and string types. Lengths can be specified as a comma-separated list of integers enclosed in curly braces, or as a single integer.",
```

karmacoma-eth: does it really work with curly braces? in my shell it has a special meaning and does substitutions:

```sh
$ echo a={1,2}
a=1 a=2
```

src/halmos/config.py
```diff
@@ -231,15 +243,22 @@ class Config:
...
+        help="set default lengths for dynamic-sized arrays (excluding bytes and string) not specified in --array-lengths",
+        global_default="0,1,2",
+        metavar="LENGTH1,LENGTH2,...",
+        action=ParseCSV,
+    )
+
     default_bytes_lengths: str = arg(
-        help="set the default length candidates for bytes and string not specified in --array-lengths",
-        global_default="0,1024,65",  # 65 is ECDSA signature size
+        help="set default lengths for bytes and string not specified in --array-lengths",
```
karmacoma-eth, line 260:
```suggestion
        help="set default lengths for bytes and string types not specified in --array-lengths",
```

karmacoma-eth (approved):

src/halmos/config.py
daejunpark: um, you can escape curly braces in various ways, e.g., `echo 'a={1,2}'`

src/halmos/config.py
daejunpark: we do expect whitespaces, like: `halmos --array-lengths "x={1,   2}"`

daejunpark:
there's a multithreading-related issue with pytest in ci, which is orthogonal to this pr. merging this despite some test failures. trying to fix the flakiness separately.

PR #385, opened 2024-10-03
feat: remove test parallel mode and use process pool for solvers

daejunpark:


daejunpark:
when using process pool for solvers, snekmate erc721 test experienced performance overhead due to weird waiting schedule for process pool. further investigation is needed. closing for now.

PR #387, opened 2024-10-07
Update examples/README.md

daejunpark:
adds morpho blue and snekmate tests

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -3,26 +3,32 @@
 #### Usage Examples
 
 - [Simple examples](simple/README.md)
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -2,10 +2,10 @@
 
 #### Usage Examples
 
-- [Simple examples](simple/README.md)
-- [ERC20](tokens/ERC20/): verifies OpenZeppelin, Solady, Solmate ERC20 tokens, and CurveTokenV3.
+- [Simple examples](simple/test/)
```
karmacoma-eth, line 5: we're not doing the README there anymore? we could also move the README at simple/test/README.md, so we would see the test files and the explainer

karmacoma-eth (approved):

examples/README.md
daejunpark: the current readme is outdated, and needs to be updated to cover more examples there. we can update the link to the readme once it is updated.

PR #388, opened 2024-10-07
fix: exclude test contract from symbolic calls

daejunpark:
previously, all deployed contracts, including the test contract, were considered for calls to symbolic addresses. however, the test contract is typically not expected to be involved in such symbolic calls. moreover, recursively calling back to the test contract could lead to infinite looping behavior.

now, the test contract is excluded from symbolic calls to prevent this issue.

### DIFF
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -213,7 +213,7 @@ class Prank:
 def symbolic_storage(ex, arg, sevm, stack, step_id):
     account = uint160(arg.get_word(4))
     account_alias = sevm.resolve_address_alias(
-        ex, account, stack, step_id, branching=False
+        ex, account, stack, step_id, allow_branching=False
     )
 
     if account_alias is None:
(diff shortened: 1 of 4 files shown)


Elyx0:
I'm new to the symbolic logic, sometimes in hack reproductions we use the foundry test contract itself as callback receiver like https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-09/Bankroll_exp.sol if it doesn't break that behavior that would be fine. Or make it clear that we have to deploy a specific Attacker contract in the setup beforehands to use symbolic functionality

daejunpark:
> I'm new to the symbolic logic, sometimes in hack reproductions we use the foundry test contract itself as callback receiver like https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-09/Bankroll_exp.sol if it doesn't break that behavior that would be fine. Or make it clear that we have to deploy a specific Attacker contract in the setup beforehands to use symbolic functionality

@Elyx0 thanks for flagging! in your particular example, no addresses are symbolic, so this change won't have an impact. however, if you have a symbolic address, e.g., one passed as a test function parameter, or created using svm.createAddress() cheatcode, then you would need to deploy separate attacker contracts in setup(). otherwise, allowing callbacks to the test contract could lead to infinite loops.

we're merging this, but please let us know if it doesn't work for your use cases!

PR #389, opened 2024-10-09
fix: reset z3 solver objects

daejunpark:
this pr addresses several issues related to flakiness in ci runs, including:

- **solver resets**: using multiple z3 solver objects, especially with the global context, may cause segfault after garbage collection of those z3 objects. this has led to flakiness in ci runs. this pr ensures that solver objects are properly reset at the end of their use.
  - this pr subsumes #390, which simplifies the implementation of proper solver resets.

- **solver contexts**: where possible, certain independent solvers are now created with their own context (separate from the global context) to further reduce the risk of conflicts.

- **symbol names**: moreover, to avoid name clashes, z3 symbol names now include a uuid to ensure uniqueness.

- **python garbage collection**: there was an issue at the python level with automatic garbage collection for cyclic objects, which sometimes incorrectly collected z3 objects that were still in use. to address this, a new `--disable-gc` flag has been added and is used in ci tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
       matrix:
         os: ["macos-latest", "ubuntu-latest", "windows-latest"]
         python-version: ["3.11", "3.12"]
+        storage-layout: ["solidity", "generic"]
+        cache-solver: ["", "--cache-solver"]
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -27,15 +27,14 @@ jobs:
...
-        parallel: ["", "--test-parallel"]
         storage-layout: ["solidity", "generic"]
         cache-solver: ["", "--cache-solver"]
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          submodules: false
+          submodules: recursive
```
daejunpark, line 37: submodules are checked out during pytest anyway

requirements-dev.txt
```diff
@@ -1,4 +1,3 @@
 black
 pre-commit
 pytest
-pytest-xdist
```
daejunpark, line 4: to ensure xdist is not enabled

src/halmos/utils.py
```diff
@@ -221,9 +221,6 @@ def create_solver(logic="QF_AUFBV", ctx=None, timeout=0, max_memory=0):
     # QF_AUFBV: quantifier-free bitvector + array theory: https://smtlib.cs.uiowa.edu/logics.shtml
     solver = SolverFor(logic, ctx=ctx)
 
-    # reset any remaining solver states from the default context
-    solver.reset()
-
```
daejunpark, line 226: now solvers are reset at the end of their life. this differs from the previous approach, as there are two active solvers: one for setup, another for each test.

src/halmos/__main__.py
```diff
@@ -504,11 +504,31 @@ def setup(
...
+        # convert model into string to avoid pickling errors for z3 (ctypes) objects containing pointers
+        self.model = (
+            to_str_model(model, args.print_full_model)
+            if isinstance(model, ModelRef)
+            else model
+        )
+        self.is_valid = is_model_valid(model)
+
+    def __str__(self) -> str:
+        if isinstance(self.model, str):
```
karmacoma-eth, line 521:
```suggestion
        # expected to be a filename
        if isinstance(self.model, str):
```

.github/workflows/test.yml
karmacoma-eth: IIRC it is faster to let foundry do a sparse checkout of the missing libraries rather than having the action do a git checkout of the submodules

daejunpark: oh i see. i can revert this

karmacoma-eth (approved):

PR #390, opened 2024-10-09
feat: remove test parallel mode

daejunpark:
the test parallel mode is not much used and may lead to performance degradation when used together parallel solving (which is now always enabled). this is due to too many parallel cores running (# of parallel tests * # of parallel solvers). this issue is expected to worsen with the introduction of portfolio solving later.

this pr removes the test parallel feature.


daejunpark:
subsumed by #389

Issue #229, opened 2023-12-14
Better handling of concrete keccaks


karmacoma-eth:
```solidity

// SPDX-License-Identifier: MIT
pragma solidity ^0.8.15;
import "forge-std/Test.sol";

contract Test49 is Test {
    function test_concreteKeccak_up() external {
        bytes32 hash = keccak256(abi.encodePacked(uint256(3)));
        uint256 bit = uint256(hash) & 1;

        if (uint256(hash) & 1 == 0) {
            console2.log("even");
        } else {
            console2.log("odd");
        }
    }

    function test_concreteKeccak_lookup() external {
        bytes32 hash = keccak256(abi.encodePacked(uint256(3)));
        uint256 bit = uint256(hash) & 1;

        string[] memory x = new string[](2);
        x[0] = "even";
        x[1] = "odd";

        console2.log(x[bit]);
    }
}
```

```
Running 2 tests for test/49_concreteKeccak.t.sol:Test49
[FAIL] test_concreteKeccak_lookup() (paths: 0/1, time: 0.02s, bounds: [])
WARNING:Halmos:Encountered symbolic memory offset: 224 + 32*Concat(0, Extract(0, 0, sha3_256(3)))
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
[console.log] even
[console.log] odd
[PASS] test_concreteKeccak_up() (paths: 2/2, time: 0.03s, bounds: [])
Symbolic test result: 1 passed; 1 failed; time: 0.06s
```

ideally we don't want the symbolic memory offset error and only a single console.log

karmacoma-eth:
Related to https://github.com/a16z/halmos/issues/212 and https://github.com/a16z/halmos/issues/214

karmacoma-eth:
Another example reported by @hsellami

PR #391, opened 2024-10-10
feat: handle concrete keccaks concretely (WIP)

karmacoma-eth:
fixes https://github.com/a16z/halmos/issues/229

### DIFF
--- a/src/halmos/utils.py
+++ b/src/halmos/utils.py
@@ -21,6 +21,7 @@ from z3 import (
     If,
     Not,
     SignExt,
+    Solver,
     SolverFor,
     ZeroExt,
     eq,
@@ -30,6 +31,7 @@ from z3 import (
     is_bv,
     is_bv_value,
     is_not,
+    sat,
     simplify,
 )
 
@@ -340,7 +342,9 @@ def unbox_int(x: Any) -> Any:
     return x
 
 
-def int_of(x: Any, err: str = "expected concrete value but got") -> int:
+def int_of(
+    x: Any, err: str = "expected concrete value but got", solver: Solver = None
+) -> int:
     """
     Converts int-like objects to int or raises NotConcreteError
     """
@@ -348,6 +352,14 @@ def int_of(x: Any, err: str = "expected concrete value but got") -> int:
 
     if isinstance(res, int):
         return res
+
+    # if the value is still symbolic at this point, check if it can be evaluated
+    # concretely with the current solver state
+    if solver is not None and solver.check() == sat:
+        m = solver.model()
+        res = m.evaluate(x)
+        if is_bv_value(res):
+            return res.as_long()
 
     raise NotConcreteError(f"{err}: {x}")
 
--- a/src/halmos/utils.py
+++ b/src/halmos/utils.py
@@ -22,7 +22,6 @@ from z3 import (
     If,
     Not,
     SignExt,
-    Solver,
     SolverFor,
     ZeroExt,
     eq,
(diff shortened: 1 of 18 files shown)


karmacoma-eth:
need to investigate some test failures, but want to open this for feedback about the general approach. It does work with the test examples in the linked issue!

karmacoma-eth:
cc @elyx0 -- your example should work with this (experimental)

Elyx0:
Awesome! I guess I'll wait for it to hit my pipx upgrade once the draft passes
```
pipx upgrade halmos
halmos is already at latest version 0.2.0 (location:  .local/pipx/venvs/halmos)
```

src/halmos/utils.py
```diff
@@ -349,6 +353,14 @@ def int_of(x: Any, err: str = "expected concrete value but got") -> int:
...
         return res
 
+    # if the value is still symbolic at this point, check if it can be evaluated
+    # concretely with the current solver state
+    if solver is not None and solver.check() == sat:
+        m = solver.model()
+        res = m.evaluate(x)
+        if is_bv_value(res):
+            return res.as_long()
+
```
daejunpark, line 363: could you elaborate on the context of this?

karmacoma-eth: see the explanation in sha3_expr:

            # we have a choice here: return the concrete hash value or the symbol
            # - if we return the value, we can directly process expressions
            # - if we return the symbol, we can use the extra information to help
            #   with things like recovering storage base slots
            #
            # for now, we return the symbol, and if we need a concrete value, we can
            # lazily evaluate it at the point of use
            #
            # see: halmos.utils:int_of

returning the concrete hash makes evaluating expressions easy, but breaks things that expect to parse symbols like the storage processing code. Returning symbols like `f_sha3_N(x)` keeps the storage processing code happy, but breaks if you want to do something like `mload(1 + f_sha3_N(x))`

so the middle ground I went with is to keep the expression symbolic, but when we attempt to get a concrete value out of it, we need a way to replace `f_sha3_256(0)` with the actual value. Using `model.evaluate` seems to do the trick efficiently

karmacoma-eth: the problem is that we need a solver to do the evaluation (instead of just erroring out), and so I shoved one in `State` which is arguably not great.

karmacoma-eth: specifically it lets us support `check_concrete_keccak_memory_lookup`

karmacoma-eth: ok it is indeed not sound:

```
>>> s = Solver()
>>> s.add(x < 3)
>>> s.check()
sat
>>> m = s.model()
>>> m.evaluate(x)
2
>>> m.evaluate(x + 1)
3
```

looks like we need to substitute the symbol with its value then

daejunpark:
trying to narrow the scope into adding an equality between concrete keccak term and its hash value. this alone can resolve various false counterexample issues. supporting the case where concrete keccak is used as index for memory or calldata, or size of dynamic parameters, has been left as future work.

todo: significant performance improvements were observed in some tests (including snekmate). need to check whether they are benign or due to incorrectly omitting paths.

daejunpark:
@karmacoma-eth could you please add @Elyx0's example as a test?

daejunpark (approved):
we can merge once the docker build is fixed. thanks!

karmacoma-eth:
the docker build requires publishing a new version of halmos-builder:latest, so we need to merge this to main first. Will fix separately

karmacoma-eth:
> the docker build requires publishing a new version of `halmos-builder:latest`, so we need to merge this to main first. Will fix separately

fixed now that the new halmos-builder has been published:
https://github.com/a16z/halmos/actions/runs/11330524487/job/31508676038

PR #395, opened 2024-10-14
👷 Pin Vyper Temporarily to Specific Commit

pcaversaccio:
The Vyper commit https://github.com/vyperlang/vyper/commit/6843e7915729f3a3ea0d8c765dffa52033f5818e introduced an unexpected compilation regression. Until it's fixed, we pin the Vyper version used in CI for the snekmate tests to commit: https://github.com/vyperlang/vyper/commit/6606ded787184be82de9ea05aeb1b61bcd219d6a.


karmacoma-eth (approved):
thanks for the fix! 🫡

PR #401, opened 2024-11-06
fix: improve error-handling for unsupported symbolic names in cheatcodes

daejunpark:
provide a better error message when symbolic string terms are passed to cheatcodes that require concrete string values.


karmacoma-eth (approved):
👍

PR #404, opened 2024-11-08
update solvers and halmos-builder dockerfiles

karmacoma-eth:
there is an issue where jsi exits because in the docker environment, bash (i.e. jsi's parent) has PID 1

next steps:
- fix jsi and release a jsi update
- rebuild `solvers` and `halmos-builder` with that new version
- the new dockerfiles should not require new changes

with these images, we can run `foundryup` in CI to always run with the latest foundry


.github/workflows/test-external.yml
```diff
@@ -89,11 +81,19 @@ jobs:
...
+
+          # commit the result back to halmos-image
+          docker commit halmos-image-foundryup halmos-image
+
       - name: Install Vyper
         if: ${{ matrix.project.dir == 'snekmate' }}
         run: |
-          docker run --name halmos-vyper --entrypoint uv halmos-image pip install git+https://github.com/vyperlang/vyper@master
-          docker commit --change 'ENTRYPOINT ["halmos"]' halmos-vyper halmos-image
+          docker run --name halmos-image-vyper uv pip install git+https://github.com/vyperlang/vyper@master
```
daejunpark, line 95: don't we need to provide the base image name as well?

i mean, like:
```suggestion
          docker run --name halmos-image-vyper halmos-image uv pip install git+https://github.com/vyperlang/vyper@master
```

.github/workflows/test-long.yml
```diff
@@ -26,26 +26,18 @@ jobs:
...
-
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
           submodules: recursive
 
       - name: Build image
         run: |
-          docker build -t halmos-image . --file packages/halmos/Dockerfile
+          docker build -t halmos-image  . --file packages/halmos/Dockerfile
```
daejunpark, line 36:
```suggestion
          docker build -t halmos-image . --file packages/halmos/Dockerfile
```

.github/workflows/test-external.yml
```diff
@@ -72,14 +72,6 @@ jobs:
...
 
     steps:
-      # TODO: remove this step once halmos-builder package is public
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
```
daejunpark, line 82: are we going to make halmos-builder public as well? if i remember correctly, docker build previously failed without this login credential.

packages/solvers/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     g++ zlib1g-dev libboost-all-dev flex bison
 
 # Install Z3
-COPY --from=ghcr.io/z3prover/z3:ubuntu-20.04-bare-z3-sha-770c51a /usr/bin/z3 /usr/local/bin/z3
+COPY --from=ghcr.io/z3prover/z3:ubuntu-20.04-bare-z3-20240922 /usr/bin/z3 /usr/local/bin/z3
```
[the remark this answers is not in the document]
karmacoma-eth: z3 --version says 4.13.1

.github/workflows/test-external.yml
karmacoma-eth: yep good catch

.github/workflows/test-external.yml
karmacoma-eth: ah that's right, I forgot about that. Let's do it!

packages/solvers/Dockerfile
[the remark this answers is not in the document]
daejunpark: hmm, no release tag for 4.13.1: https://github.com/Z3Prover/z3/releases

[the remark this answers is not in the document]
daejunpark: it seems that the z3 docker images aren't synced with release versions. no latest version is available as a package:
https://github.com/z3prover/z3/pkgs/container/z3/versions

i think it would be more reliable to install from the github release files, as we do for yices and cvc5.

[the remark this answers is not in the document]
karmacoma-eth: in [9ad0087](https://github.com/a16z/halmos/pull/404/commits/9ad0087bc6a5446c8b520abcad1461659ff60309)

PR #406, opened 2024-11-12
perf: improve hash injectivity constraints

daejunpark:
Previously, we formulated the hash injectivity axiom as `hash(x) == hash(y) ==> x == y`, which required generating local constraints for all combinations of hash inputs, resulting in O(n^2).

Now, we reformulate the injectivity axiom as "there exists f, such that `f(hash(x)) == x`," which requires only O(n) constraints, each of which is independent from other constraints.

### DIFF
--- a/src/halmos/utils.py
+++ b/src/halmos/utils.py
@@ -109,6 +109,14 @@ def f_sha3_name(bitsize: int) -> str:
     return f"f_sha3_{bitsize}"
 
 
+def f_inv_sha3_name(bitsize: int) -> str:
+    return f"f_inv_sha3_{bitsize}"
+
+
+# TODO: explore the impact of using a smaller bitsize for the range sort
+f_inv_sha3_size = Function("f_inv_sha3_size", BitVecSort256, BitVecSort256)
+
+
 f_sha3_0_name = f_sha3_name(0)
 f_sha3_256_name = f_sha3_name(256)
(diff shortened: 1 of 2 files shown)


src/halmos/utils.py
```diff
@@ -109,6 +109,14 @@ def f_sha3_name(bitsize: int) -> str:
     return f"f_sha3_{bitsize}"
 
 
+def f_inv_sha3_name(bitsize: int) -> str:
+    return f"f_inv_sha3_{bitsize}"
+
+
+# TODO: explore the impact of using a smaller bitsize for the range sort
+f_inv_sha3_size = Function("f_inv_sha3_size", BitVecSort256, BitVecSort256)
```
karmacoma-eth, line 117: BitVecSort32 should work for the input size, right?

karmacoma-eth (approved):
very nice! excited to give it a try. I can see this helping with storage-heavy tests

src/halmos/utils.py
daejunpark: does the max size of byte arrays that can appear during execution not exceed 2^32 bits (2^29 bytes)?

karmacoma-eth: that's definitely not realistic in the current EVM execution context. 2^20 is already huge (1MB)

PR #417, opened 2024-11-18
feat: display spinning progress

daejunpark:
display a spinning progress for path exploration and assertion solving:

Screenshot 2024-12-03 at 5 06 53 PM

Screenshot 2024-12-03 at 5 07 38 PM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,8 @@ dependencies = [
     "sortedcontainers>=2.4.0",
     "toml>=0.10.2",
     "z3-solver==4.12.6.0",
-    "eth_hash[pysha3]>=0.7.0"
+    "eth_hash[pysha3]>=0.7.0",
+    "rich>=13.8.1"
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "toml>=0.10.2",
     "z3-solver==4.12.6.0",
     "eth_hash[pysha3]>=0.7.0",
-    "rich>=13.8.1"
+    "rich>=13.9.4"
 ]
 dynamic = ["version"]
 
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -745,7 +745,7 @@ def run(
                 if done == total:
                     break
                 status.update(f"solving: {done} / {total}")
-                time.sleep(1)
+                time.sleep(0.1)
 
     if args.early_exit:
         thread_pool.shutdown(wait=False, cancel_futures=True)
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -25,7 +25,8 @@ dependencies = [
     "sortedcontainers>=2.4.0",
     "toml>=0.10.2",
     "z3-solver==4.12.6.0",
-    "eth_hash[pysha3]>=0.7.0"
+    "eth_hash[pysha3]>=0.7.0",
+    "rich>=13.8.1"
```
daejunpark, line 29: @karmacoma-eth i copied the minimum version requirement from jsi. could you confirm it's still applicable here?

daejunpark:
@karmacoma-eth now ready to review.

src/halmos/sevm.py
```diff
@@ -2651,6 +2666,19 @@ def finalize(ex: Exec):
                 prev_step_id: int = item.step
                 step_id += 1
 
+                # display progress
+                if step_id % PULSE_INTERVAL == 0:
+                    elapsed = timer() - stack.start_time
+                    speed = step_id / elapsed
+
+                    # hh:mm:ss
+                    elapsed_fmt = str(timedelta(seconds=elapsed)).split(".")[0]
```
daejunpark, line 2675: Q: is there a better way to format the elapsed time as hh:mm:ss?

src/halmos/__main__.py
```diff
@@ -733,14 +734,20 @@ def future_callback(future_model):
...
+    with Status("solving:") as status:
+        while True:
+            if args.early_exit and len(counterexamples) > 0:
+                break
+            done = sum(fm.done() for fm in future_models)
+            total = len(future_models)
+            if done == total:
+                break
+            status.update(f"solving: {done} / {total}")
             time.sleep(1)
```
daejunpark, line 747: downside of this polling loop:  the assertion solving phase takes a minimum of 1s, even if the solver finishes in less than 1s.  wouldn't be a big deal in most cases, but this causes the regression test set to take longer, because it involves many small tests.

karmacoma-eth (approved):
minor comments but overall love seeing the live updates during a long run 👌

src/halmos/__main__.py
```diff
@@ -733,14 +734,21 @@ def future_callback(future_model):
...
+        with Status("solving:") as status:
+            while True:
+                if args.early_exit and len(counterexamples) > 0:
+                    break
+                done = sum(fm.done() for fm in future_models)
+                total = len(future_models)
+                if done == total:
+                    break
+                status.update(f"solving: {done} / {total}")
+                time.sleep(1)
```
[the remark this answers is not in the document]
daejunpark: i wasn't sure if polling it every sub-second would be okay performance-wise. would 0.1s be acceptable?

PR #418, opened 2024-11-20
feat: improve logging

daejunpark:
improve logging with using logger and rich handler:
- filter out duplicate debugging messages
- avoid local conditional debugging


src/halmos/warnings.py
```diff
@@ -28,4 +81,4 @@ def url(self) -> str:
 
 
 def warn_code(error_code: ErrorCode, msg: str):
-    logger.warning(f"{color_warn(msg)}\n(see {error_code.url()})")
+    logger.warning(f"{msg}\n(see {error_code.url()})")
```
daejunpark, line 84: this is to avoid circular dependency. note that now warnings are clearly marked with a level column and indentation, thanks to the rich handler.

daejunpark: e.g:
Screenshot 2024-11-20 at 4 41 52 PM

Screenshot 2024-11-20 at 4 41 26 PM

daejunpark:
@karmacoma-eth this is now ready to review

src/halmos/calldata.py
```diff
@@ -139,10 +140,9 @@ def get_dyn_sizes(self, name: str, typ: Type) -> tuple[list[int], BitVecRef]:
                 if isinstance(typ, DynamicArrayType)
                 else self.args.default_bytes_lengths  # bytes or string
             )
-            if self.args.debug:
-                print(
-                    f"Warning: no size provided for {name}; default value {sizes} will be used."
-                )
+            debug_once(
+                f"Warning: no size provided for {name}; default value {sizes} will be used."
```
karmacoma-eth, line 144:
```suggestion
                f"no size provided for {name}; default value {sizes} will be used."
```

karmacoma-eth (approved):
looks nice, we should have done this sooner! i think it would make sense to rename warnings.py to something else since it now deals with all logging

(does `logging.py` conflict with the built-in `logging`? maybe just `logs.py`?)

PR #421, opened 2024-12-03
ci: add python 3.13 to test.yml

karmacoma-eth:


daejunpark (approved):

daejunpark:
should we update it in the pyproject as well?

https://github.com/a16z/halmos/blob/main/pyproject.toml#L38-L39

karmacoma-eth:
> should we update it in the pyproject as well?
> 
> https://github.com/a16z/halmos/blob/main/pyproject.toml#L38-L39

no, that's a black setting which we don't use anymore. Ruff just relies on the existing `requires-python = ">=3.11"` setting from pyproject.toml

PR #422, opened 2024-12-04
Embrace uv

karmacoma-eth:
- recommend the uv workflow in README.md and CONTRIBUTING.md
- replace requirements-dev.txt with an optional dependency group in pyproject.toml
- replace remaining mentions of black with ruff
- switch relevant workflows and dockerfiles to uv


uv.lock
```diff
@@ -0,0 +1,277 @@
...
+version = "3.16.1"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/9d/db/3ef5bb276dae18d6ec2124224403d1d67bccdbefc17af4cc8f553e341ab1/filelock-3.16.1.tar.gz", hash = "sha256:c249fbfcd5db47e5e2d6d62198e565475ee65e4831e2561c8e313fa7eb961435", size = 18037 }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/b9/f8/feced7779d755758a52d1f6635d990b8d98dc0a29fa568bbe0625f18fdf3/filelock-3.16.1-py3-none-any.whl", hash = "sha256:2082e5703d51fbf98ea75855d9d5527e33d8ff23099bec374a134febee6946b0", size = 16163 },
+]
+
+[[package]]
+name = "halmos"
+version = "0.2.2.dev5+gb9cf162.d20241204"
```
daejunpark, line 56: would this need to be updated every commit or pr, especially for ci runs?

karmacoma-eth: I don't think so -- but went with no uv.lock in git after all

karmacoma-eth:
> looks great, thanks!
> 
> two things:
> 
>     * could you check the ffi test failure in the ci?
> 
>     * could you confirm it's intentional to include uv.lock file in the repo?

failures are fixed, and we decided to not include the uv.lock file in the repo after all (so we have lower friction and commit noise at the cost fully reproducible builds)

PR #424, opened 2024-12-06
docs: Fix grammatical error getting-started.md

mdqst:
Снимок экрана 2024-12-06 в 13 27 05

The phrase "Symbolic tests looks similar to fuzz tests" was using an incorrect form of the verb "look." Since "tests" is plural, the verb should be "look" instead of "looks."

The corrected sentence now reads:
"Symbolic tests look similar to fuzz tests."


karmacoma-eth (approved):
good catch, thanks!

PR #429, opened 2024-12-20
feat: mappings from bytecode to contract name

daejunpark:


src/halmos/mapper.py
```diff
@@ -83,21 +84,123 @@ class BuildOut(metaclass=SingletonMeta):
...
+        Return the contract name and file name of the given deployed bytecode.
+        """
+        if not self._build_out_map_code:
+            self.create_build_out_map_code()
+
+        # compares the deployed bytecode with the compile-time hexcode, excluding placeholders from the comparison
+        def eq_except_placeholders(hexcode: str, placeholders):
+            last = 0
+            for start, end in placeholders:
+                if not eq_bytes(bytecode[last:start], hexcode[2 * last : 2 * start]):
```
karmacoma-eth, line 183: I see, we have to do the indexing dance because for immutables placeholders look like `0000...0000` but for libraries they look like `__$72fd9f18565b1bf49af679aa1eb458ccdd$__`

karmacoma-eth:
looks very clean to me! this more or less deprecates the existing `Mapper.add_mapping(self, mapping: ContractMappingInfo)`, right? (since each contract now has its own mapping as it is deployed)

karmacoma-eth (approved):

src/halmos/mapper.py
```diff
@@ -83,21 +84,123 @@ class BuildOut(metaclass=SingletonMeta):
...
+                hexcode = deployed["object"][2:]  # drop '0x' prefix
+                if not hexcode:
+                    continue
+
+                size = len(hexcode) // 2  # byte length
+                placeholders = self.get_placeholders(deployed)
+                code_data = (hexcode, placeholders, contract_name, filename)
+                self._build_out_map_code[size].append(code_data)
+
+    def get_by_code(self, bytecode: ForwardRef("ByteVec")) -> tuple:
```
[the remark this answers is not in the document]
daejunpark: if i remember correctly, (2) was complained by ruff, and (3) failed due to circular dependency.

src/halmos/mapper.py
daejunpark: yes exactly.

daejunpark:
> this more or less deprecates the existing `Mapper.add_mapping(self, mapping: ContractMappingInfo)`, right? (since each contract now has its own mapping as it is deployed)

it's correct that this is related to the existing Mapper mechanism, and some functionality overlaps, e.g., get_by_bytecode() used in render_trace() for create txs, which is better to be replaced by the new get_by_code().

PR #432, opened 2025-01-06
fix: progress display race condition can result in LiveError

karmacoma-eth:


src/halmos/__main__.py
```diff
@@ -753,6 +754,9 @@ def future_callback(future_model):
 
     # display assertion solving progress
     if not args.no_status or args.early_exit:
+        # creating a new live display would fail if the previous one was still active
+        rich.get_console().clear_live()
+
```
daejunpark, line 759: could you elaborate on the race condition scenarios? my goal was to ensure only a single Status instance exists at a time. is it not the case, or is the Live instance a different thing from Status?

daejunpark (approved):

src/halmos/__main__.py
daejunpark: update from offline discussion: it's not super clear, but it might be related to the use of context manager within generator.

PR #434, opened 2025-01-07
fix: not use context manager within generator

daejunpark:
the use of rich.status context manager within the sevm generator is not robust, as certain stop-iteration events may not be properly captured by the context manager's exit handler.

to address this, a status instance is now added as a field of the sevm class. it's explicitly started and stopped in the main.run function, controlled in the same way with the timer. also, now a single status instance is used throughout each test run, instead of two separate instances.

this is a followup to #432

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -608,6 +608,7 @@ def run(
 
     timer = NamedTimer("time")
     timer.create_subtimer("paths")
+    sevm.status.start()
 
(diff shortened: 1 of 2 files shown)


karmacoma-eth:
having a single Status instance should avoid issues 👌 Something about SEVM having a Status field is a bit impure but it's fine for now, I have some ideas to improve this later

karmacoma-eth (approved):

PR #436, opened 2025-01-08
feat: improve dynamic calldata handling

daejunpark:


src/halmos/sevm.py
```diff
@@ -2017,6 +2025,85 @@ def resolve_address_alias(
...
+
+        selector = arg[0:4].unwrap()
+
+        # TODO: check if calldata can be decoded. even if selector is concrete, remaining calldata may not be decodable.
+        if (
+            isinstance(selector, bytes)
+            or is_bv_value(selector)
+            or str(selector).startswith("fallback_selector_")
+        ):
+            return arg
```
daejunpark, line 2057: another issue: even if selector is symbolic, calldata may be decodable, in which case, refining calldata is not needed, and may lead to unexpected behaviors.

daejunpark:
closing, because generating calldata for all functions at once is problematic. it needs to be done separately for each function.

PR #440, opened 2025-01-10
fix: hexify for extract and const terms

daejunpark:
previously, `Extract(hi, lo, x)` is printed as `Extract(x())` by hexify.

there were two issues:
- the hi and lo arguments appear in the decl params(), but not in the app children(). params() was missing in hexify().
- constants (including variables) have no arguments, so the trailing `()` should be removed for readability.

both issues have now been fixed, and extract terms are hexified properly.


karmacoma-eth (approved):
nice fix!

PR #442, opened 2025-01-14
fix: avoid raising internal python error on large memory access

karmacoma-eth:
these can otherwise raise `OverflowError: cannot fit 'int' into an index-sized integer`

For example:

```solidity
pragma solidity ^0.8.15;

import "forge-std/Test.sol";

contract Test112 is Test, SymTest {
    uint256 MEGA_SIZE = 10**32;

    function setUp() public {
        console2.log("Test112.setUp");
    }

    // good: reverts with Panic(0x41) (too much memory requested)
    function test_megaMem_new_bytes() external {
        bytes memory data = new bytes(MEGA_SIZE);
        console2.logBytes(data);
    }

    // bad: OverflowError: cannot fit 'int' into an index-sized integer
    function test_megaMem_sha3() external {
        bytes32 hash;
        uint256 size = MEGA_SIZE;
        assembly {
            hash := keccak256(0, size)
        }
    }

    function test_megaMem_call_insize() external {
        address addr = address(this);
        uint256 value = 0;
        bool success;
        uint256 in_ptr = 0;
        uint256 in_size = MEGA_SIZE;
        uint256 out_ptr = 0;
        uint256 out_size = 0;

        assembly {
            success := call(gas(), addr, value, in_ptr, in_size, out_ptr, out_size)
        }
    }

    // PASS
    function test_megaMem_call_outsize() external {
        address addr = address(this);
        uint256 value = 0;
        bool success;
        uint256 in_ptr = 0;
        uint256 in_size = 0;
        uint256 out_ptr = 0;
        uint256 out_size = MEGA_SIZE;

        assembly {
            success := call(gas(), addr, value, in_ptr, in_size, out_ptr, out_size)
        }
    }

    function test_megaMem_return() external returns (bytes memory) {
        uint256 size = MEGA_SIZE;
        assembly {
            return(0, size)
        }
    }

    function test_megaMem_log() external {
        uint256 size = MEGA_SIZE;
        assembly {
            log0(0, size)
        }
    }
}
```

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -472,14 +472,17 @@ class State:
 
         self.stack[-(n + 1)], self.stack[-1] = self.stack[-1], self.stack[-(n + 1)]
 
-    def mloc(self, subst: dict = None) -> int:
+    def mloc(self, subst: dict = None, check_size: bool = True) -> int:
         loc: int = int_of(self.pop(), "symbolic memory offset", subst)
-        if loc > MAX_MEMORY_SIZE:
+        if check_size and loc > MAX_MEMORY_SIZE:
             raise OutOfGasError(f"memory {loc=} > MAX_MEMORY_SIZE")
         return loc
 
     def mslice(self, loc: int, size: int) -> ByteVec:
         """Wraps a memory slice read with a size check."""
+        if not size:
+            return ByteVec()
+
         if loc + size > MAX_MEMORY_SIZE:
             raise OutOfGasError(
                 f"memory access with {loc=} {size=} exceeds MAX_MEMORY_SIZE"
@@ -490,6 +493,10 @@ class State:
     def set_mslice(self, loc: int, data: ByteVec) -> None:
         """Wraps a memory slice write with a size check."""
         size = len(data)
+
+        if not size:
+            return
+
         if loc + size > MAX_MEMORY_SIZE:
             raise OutOfGasError(
                 f"memory write with {loc=} {size=} exceeds MAX_MEMORY_SIZE"
@@ -1196,7 +1203,7 @@ class Exec:  # an execution path
         self.balances[new_balance_var] = new_balance
 
     def sha3(self) -> None:
-        loc: int = self.int_of(self.st.pop(), "symbolic SHA3 data loc")
+        loc: int = self.mloc(check_size=False)
         size: int = self.int_of(self.st.pop(), "symbolic SHA3 data size")
         data = self.st.mslice(loc, size).unwrap() if size else b""
         sha3_image = self.sha3_data(data)
@@ -1368,8 +1375,10 @@ class Exec:  # an execution path
 
         return (creation_hexcode, deployed_hexcode)
 
-    def mloc(self) -> int:
-        return self.st.mloc(self.path.concretization.substitution)
+    def mloc(self, check_size: bool = True) -> int:
+        return self.st.mloc(
+            self.path.concretization.substitution, check_size=check_size
+        )
 
     def ret(self) -> ByteVec:
         return self.st.ret(self.path.concretization.substitution)
@@ -2100,10 +2109,10 @@ class SEVM:
         to = uint160(ex.st.pop())
         fund = ZERO if op in [EVM.STATICCALL, EVM.DELEGATECALL] else ex.st.pop()
 
-        arg_loc: int = ex.mloc()
+        arg_loc: int = ex.mloc(check_size=False)
         arg_size: int = ex.int_of(ex.st.pop(), "symbolic CALL input data size")
 
-        ret_loc: int = ex.mloc()
+        ret_loc: int = ex.mloc(check_size=False)
         ret_size: int = ex.int_of(ex.st.pop(), "symbolic CALL return data size")
 
         if not arg_size >= 0:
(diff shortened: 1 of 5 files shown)


karmacoma-eth:
h/t @m4k2

src/halmos/sevm.py
```diff
@@ -3106,15 +3122,9 @@ def finalize(ex: Exec):
...
-                        dst_end_loc = dest_offset + size
-
-                        if max(src_end_loc, dst_end_loc) > MAX_MEMORY_SIZE:
-                            raise HalmosException("MCOPY > MAX_MEMORY_SIZE")
-
-                        data = ex.st.memory.slice(src_offset, src_end_loc)
-                        ex.st.memory.set_slice(dest_offset, dst_end_loc, data)
+                    if size:
+                        data = ex.st.mslice(src_offset, size)
+                        ex.st.memory.set_slice(dest_offset, dest_offset + size, data)
```
daejunpark, line 3127: previously, both src and dst were checked, but now only src is checked. is it intentional, or is it implicitly checked during set_slice?

daejunpark:
awesome, could you also add these nice examples as tests?

src/halmos/sevm.py
karmacoma-eth: you're right that's an oversight, I was focused on the case where size is huge, but dest_offset could also be huge

karmacoma-eth:
> awesome, could you also add these nice examples as tests?

do you think they fit as unit tests? The problem is that in both cases we get an error in both cases (but a different flavor), so the returncode is just 1 anyway right?

karmacoma-eth:
> > > awesome, could you also add these nice examples as tests?
> > 
> > 
> > do you think they fit as unit tests? The problem is that in both cases we get an error in both cases (but a different flavor), so the returncode is just 1 anyway right?
> 
> actually I think I have a decent pattern to test for this:
> 
> ```solidity
>     function test_megaMem_sha3(bool coinflip) external {
>         bytes32 hash;
>         uint256 size = coinflip ? MEGA_SIZE : 32;
> 
>         assembly {
>             hash := keccak256(0, size)
>         }
>     }
> ```
> 
> ```
> # before
> [ERROR] test_megaMem_sha3(bool)
> ERROR    OverflowError: cannot fit 'int' into an index-sized integer  
> 
> # after
> [PASS] test_megaMem_sha3(bool) (paths: 3, time: 0.04s (paths: 0.04s, models: 0.00s), bounds: [])
> ```

in 0d2aab3762f9197c3b68ff21f7d36c88912b53ff

daejunpark (approved):
great, thanks for thoughtfully handling all the various cases!

src/halmos/sevm.py
```diff
@@ -2302,24 +2342,21 @@ def call_unknown() -> None:
...
                 if effective_ret_size > 0:
-                    end_loc = ret_loc + effective_ret_size
-                    if end_loc > MAX_MEMORY_SIZE:
-                        raise HalmosException("returned data exceeds MAX_MEMORY_SIZE")
-
-                    new_ex.st.memory.set_slice(ret_loc, end_loc, ret_)
-
-                if not isinstance(ret_, ByteVec):
-                    raise HalmosException(f"Invalid return value: {ret_}")
+                    data = ret_.slice(0, effective_ret_size)
```
daejunpark, line 2358: if i understand correctly, this needs the full scan even if `effective_ret_size == len(ret_)`. could you make sure slice() returns immediately in such case?

src/halmos/sevm.py
```diff
@@ -463,18 +472,43 @@ def swap(self, n: int) -> None:
...
+        """Wraps a memory slice read with a size check."""
+        if not size:
+            return ByteVec()
+
+        if loc + size > MAX_MEMORY_SIZE:
+            raise OutOfGasError(
+                f"memory access with {loc=} {size=} exceeds MAX_MEMORY_SIZE"
+            )
+
+        return self.memory.slice(start=loc, stop=loc + size)
```
daejunpark, line 491: you may want to factor out `loc + size` to avoid computing them twice.

src/halmos/sevm.py
```diff
@@ -463,18 +472,43 @@ def swap(self, n: int) -> None:
...
+        if not size:
+            return
+
+        if loc + size > MAX_MEMORY_SIZE:
+            raise OutOfGasError(
+                f"memory write with {loc=} {size=} exceeds MAX_MEMORY_SIZE"
+            )
+
+        self.memory.set_slice(start=loc, stop=loc + size, value=data)
+
```
daejunpark, line 506: similarly, `loc + size` computed twice.

src/halmos/sevm.py
```diff
@@ -2920,22 +2957,16 @@ def finalize(ex: Exec):
...
                                 f"EXTCODECOPY: unknown address {hexify(account)} "
                                 "is assumed to have empty bytecode"
                             )
 
                         account_code: Contract = ex.code.get(account_alias) or ByteVec()
-                        codeslice: ByteVec = account_code._code.slice(
-                            offset, offset + size
-                        )
-                        ex.st.memory.set_slice(loc, end_loc, codeslice)
+                        codeslice: ByteVec = account_code.slice(offset, size)
```
daejunpark, line 2968: oh good catch!

daejunpark (approved):

src/halmos/sevm.py
karmacoma-eth: good point, let me look into it

src/halmos/sevm.py
karmacoma-eth: dst is now checked in set_mslice

src/halmos/sevm.py
karmacoma-eth: in 778fbde

src/halmos/sevm.py
karmacoma-eth: in 48424ca

daejunpark: https://github.com/a16z/halmos/commit/48424caf7a65ef1c69bd9d3327fc3d381eeda5c8 looks good. but it addressed the same issue in a different location, leaving this one unaddressed. alternatively, you might consider improving the bytevec.slice(0, size) function to immediately return when size is the full length?

karmacoma-eth: > you might consider improving the bytevec.slice(0, size) function to immediately return when size is the full length?

that may be a little tricky to handle properly, because bytevecs are mutable. So in principle we need to return a copy, but in this case we can skip the copy because we just blit it in memory and we know we're not going to modify the returndata any further

karmacoma-eth: I'll handle it at this callsite and we can think about reducing copies further or introducing immutable bytevecs in a separate PR

karmacoma-eth: sorry for the confusion, I refactored both places to call the same function: ebc6da6

PR #443, opened 2025-01-14
feat: shadow calldata

daejunpark:
in #436, calldata generation is done at CALL, generating calldata for all functions at once, which leads to issues.

in this pr, calldata generation is done at CALLDATALOAD, where calldata is generated for each function.


daejunpark:
this approach is still problematic. when generating shadow calldata, it fails to account for existing partial constraints on the original calldata, resulting in invalid shadow calldata. solving this issue seems to need proper sequence reasoning. closing.

PR #445, opened 2025-01-21
fix: check infeasible errors

daejunpark:
infeasible (halmos) errors may be reported if the branching timeout is not enough to prune infeasible error paths during path exploration. this pr adds checks to ensure error conditions are feasible.


src/halmos/__main__.py
```diff
@@ -724,9 +724,12 @@ def future_callback(future_model):
             future_models.append(future_model)
 
         elif ex.context.is_stuck():
-            stuck.append((idx, ex, ex.context.get_stuck_reason()))
-            if args.print_blocked_states:
-                traces[idx] = f"{hexify(ex.path)}\n{rendered_trace(ex.context)}"
+            debug(f"Potential error path (id: {idx+1})")
+            res, _, _ = solve(ex.path.to_smt2(args), args)
```
daejunpark, line 728: performance concern: this may block the path exploration thread.

an alternative approach is to add this into the assertion-solving thread pool, but it requires further code refactoring. not sure it's worth it at this point.

0xkarmacoma: yeah that would be a big change, and would break some assumptions in the assertion-solving thread pool (sincewe collect/triage/print counterexamples). If necessary we could add a separation branch-solving thread pool, but I don't think it would interact well with the z3 context push/pop mechanism (can't really keep exploring another branch until this one is done)

0xkarmacoma (approved):
good trade-off IMO 👌

Issue #451, opened 2025-01-30
bools on the stack can be mixed with binary operations


0xkarmacoma:
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.15;

import "forge-std/Test.sol";
import {console2} from "forge-std/console2.sol";

import {SymTest} from "halmos-cheatcodes/SymTest.sol";

contract Test117 is Test, SymTest {
    function setUp() public {
        console2.log("Test117.setUp");
    }

    function test_shiftByBool(uint256 x) external {
        uint256 y;
        assembly {
            let cond := eq(x, 42)
            y := shr(x, cond)
        }
        assertNotEq(y, 42);
    }
}
```

```
$ halmos --function test_shiftByBool

ERROR    Z3Exception: sort mismatch                                                                                                                                                                                                                                                                                     
Traceback (most recent call last):
  File "/Users/karma/projects/halmos/src/halmos/__main__.py", line 773, in run_contract
    test_result = run_test(test_ctx)
                  ^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/src/halmos/__main__.py", line 523, in run_test
    for path_id, ex in enumerate(exs):
  File "/Users/karma/projects/halmos/src/halmos/sevm.py", line 2943, in run
    ex.st.push(LShR(ex.st.pop(), w))  # bvlshr
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/.venv/lib/python3.12/site-packages/z3/z3.py", line 4373, in LShR
    a, b = _coerce_exprs(a, b)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/.venv/lib/python3.12/site-packages/z3/z3.py", line 1235, in _coerce_exprs
    s = _coerce_expr_merge(s, b)
        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karma/projects/halmos/.venv/lib/python3.12/site-packages/z3/z3.py", line 1215, in _coerce_expr_merge
    _z3_assert(False, "sort mismatch")
  File "/Users/karma/projects/halmos/.venv/lib/python3.12/site-packages/z3/z3.py", line 107, in _z3_assert
    raise Z3Exception(msg)
z3.z3types.Z3Exception: sort mismatch
```

0xkarmacoma:
[...]

PR #452, opened 2025-01-31
fix: avoid sort mismatches bv <> bool

0xkarmacoma:
fixes https://github.com/a16z/halmos/issues/451

a lot of places actually already had a conversion for bool in place, using `b2i(ex.st.pop())`

this introduces `st.popi()` which pops and does the `b2i` conversion, and spreads it to more instructions

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -492,6 +492,10 @@ class State:
             raise StackUnderflowError()
         return self.stack.pop()
 
+    def popi(self) -> Word:
+        """The stack can contain BitVecs or Bools -- this function converts Bools to BitVecs"""
+        return b2i(self.pop())
+
     def peek(self, n: int = 1) -> Word:
         return self.stack[-n]
 
@@ -508,7 +512,7 @@ class State:
         self.stack[-(n + 1)], self.stack[-1] = self.stack[-1], self.stack[-(n + 1)]
 
     def mloc(self, subst: dict = None, check_size: bool = True) -> int:
-        loc: int = int_of(self.pop(), "symbolic memory offset", subst)
+        loc: int = int_of(self.popi(), "symbolic memory offset", subst)
         if check_size and loc > MAX_MEMORY_SIZE:
             raise OutOfGasError(f"memory {loc=} > MAX_MEMORY_SIZE")
         return loc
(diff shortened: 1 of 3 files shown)


src/halmos/sevm.py
```diff
@@ -1733,12 +1737,8 @@ def bitwise(op, x: Word, y: Word) -> Word:
             return x ^ y  # bvxor
         else:
             raise ValueError(op, x, y)
-    elif is_bool(x) and is_bv(y):
-        return bitwise(op, If(x, ONE, ZERO), y)
-    elif is_bv(x) and is_bool(y):
-        return bitwise(op, x, If(y, ONE, ZERO))
     else:
-        raise ValueError(op, x, y)
+        return bitwise(op, b2i(x), b2i(y))
```
daejunpark, line 1741: this would result in infinite recursion if x or y is neither bool or bv.  it doesn't seem to happen in the current code, but it might be in the future.  adding an explicit check would be safer.

daejunpark:
> this introduces `st.popi()` which pops and does the `b2i` conversion, and spreads it to more instructions

so the idea is to use `popi()` only when an integer is required, right?  that's good approach, as it avoids unnecessary conversions.

tests/expected/all.json
```diff
@@ -3093,6 +3093,242 @@
...
+            {
+                "name": "check_xor(uint256)",
+                "exitcode": 0,
+                "num_models": 0,
+                "models": null,
+                "num_paths": null,
+                "time": null,
+                "num_bounded_loops": null
+            }
+        ],
```
daejunpark, line 3331: did you forget to add the new test file?

0xkarmacoma: yes I did 🤦‍♂️

0xkarmacoma:
> > this introduces `st.popi()` which pops and does the `b2i` conversion, and spreads it to more instructions
> 
> so the idea is to use `popi()` only when an integer is required, right? that's good approach, as it avoids unnecessary conversions.

that's right. Some instructions like AND/OR may want to handle bool vs bv themselves, but for instructions like ADD/MUL/SHR etc, I think it makes sense that when they pop an item from the stack they can be sure it has been converted to bv

tests/expected/all.json
0xkarmacoma: in [111f8bb](https://github.com/a16z/halmos/pull/452/commits/111f8bb1688975228251ff5e06f30e1d361a589f)

src/halmos/sevm.py
daejunpark: that's good idea too!

daejunpark (approved):

PR #454, opened 2025-02-06
feat: invariant testing

daejunpark:
mvp for invariant testing.

further improvements and features needed (in seperate prs):
- refactor invariant test result display logic
   - https://github.com/a16z/halmos/pull/454#discussion_r1978329493
   - https://github.com/a16z/halmos/pull/454#discussion_r1977926911
   - https://github.com/a16z/halmos/pull/454#discussion_r1977944178
- support for target control, e.g., targetContract(), excludeContract()
   - https://github.com/a16z/halmos/pull/454#discussion_r1977922459
- support for sender control
   - https://github.com/a16z/halmos/pull/454#discussion_r1978356009
- fix duplicate test config parsing
   - https://github.com/a16z/halmos/pull/454#pullrequestreview-2655548277

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -509,6 +509,7 @@ def run_single_invariant_step(
                     return next_exs, funsigs
 
                 # note: any state changes made during invariant checking are excluded, to reduce path condition complexity
+                post_ex.path_slice()
                 next_exs.append(post_ex)
 
     return next_exs, funsigs
@@ -595,20 +596,20 @@ def run_target_contract(ctx, ex, addr) -> list[Exec]:
             post_ex.context = deepcopy(ex.context)
             post_ex.context.trace.append(subcall)
 
-            var_set = get_vars(post_ex.balance)
-            for _addr, _storage in post_ex.storage.items():
-                var_set = itertools.chain(var_set, get_vars(_addr))
-                for _, _val in _storage._mapping.items():
-                    var_set = itertools.chain(var_set, get_vars(_val))
+#           var_set = get_vars(post_ex.balance)
+#           for _addr, _storage in post_ex.storage.items():
+#               var_set = itertools.chain(var_set, get_vars(_addr))
+#               for _, _val in _storage._mapping.items():
+#                   var_set = itertools.chain(var_set, get_vars(_val))
 
-            related = post_ex.path._get_related(var_set)
-            post_path = Path(solver)
-            idx = 0
-            for cond, branching in post_ex.path.conditions.items():
-                if idx in related:
-                    post_path.append(cond, branching)
-                idx += 1
-            post_ex.path = post_path
+#           related = post_ex.path._get_related(var_set)
+#           post_path = Path(solver)
+#           idx = 0
+#           for cond, branching in post_ex.path.conditions.items():
+#               if idx in related:
+#                   post_path.append(cond, branching)
+#               idx += 1
+#           post_ex.path = post_path
 
             results.append(post_ex)
 
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -403,8 +403,7 @@ def is_global_fail_set(context: CallContext) -> bool:
 
 
 def get_state_id(ex: Exec) -> bytes:
-    # todo: consider path constraints relevant to storage
-    return snapshot_state(ex, None, None, None, None).unwrap()
+    return snapshot_state(ex, include_path = True).unwrap()
 
 
 # execute invariant test functions in multiple depths
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -22,6 +22,7 @@ from z3 import (
     BitVec,
     set_option,
     unsat,
+    eq,
 )
 
 from .build import (
@@ -35,6 +36,7 @@ from .bytevec import ByteVec
 from .calldata import FunctionInfo, get_abi, mk_calldata
 from .config import Config as HalmosConfig
 from .config import arg_parser, default_config, resolve_config_files, toml_parser
+from .cheatcodes import snapshot_state
 from .constants import (
     VERBOSITY_TRACE_CONSTRUCTOR,
     VERBOSITY_TRACE_COUNTEREXAMPLE,
@@ -394,6 +396,96 @@ def is_global_fail_set(context: CallContext) -> bool:
     return hevm_fail or any(is_global_fail_set(x) for x in context.subcalls())
 
 
+
+
+def run_invariant(test_ctx):
+
+    test_result = run_test(test_ctx) # check invariant against setup state
+    if test_result.exitcode != Exitcode.PASS.value:
+        return test_result
+
+    curr = [test_ctx.setup_ex]
+
+    for i in range(2):
+        test_result, curr = run_invariant_single(test_ctx, curr)
+        if test_result.exitcode != Exitcode.PASS.value:
+            return test_result
+
+    return test_result
+
+def run_invariant_single(test_ctx, pre_exs) -> list[Exec]:
+
+    next_exs = []
+
+    for pre_ex in pre_exs:
+        pre_id = snapshot_state(pre_ex, None, None, None, None)
+
+        for addr in pre_ex.code:
+            if eq(addr, con_addr(FOUNDRY_TEST)):
+                continue
+
+            post_exs = run_target_contract(test_ctx, pre_ex, addr)
+
+            # remove non-updated states; in more general, remove visited states
+            # todo: check path feasibility
+            post_exs = filter(lambda post_ex: post_ex.output.data is not None and post_ex.output.error is None and pre_id != snapshot_state(post_ex, None, None, None, None), post_exs)
+
+            for post_ex in post_exs:
+                new_test_ctx = FunctionContext(
+                    args=test_ctx.args,
+                    info=test_ctx.fun_info,
+                    solver=mk_solver(test_ctx.args),
+                    contract_ctx=test_ctx.contract_ctx,
+                    setup_ex=post_ex,
+                )
+
+                test_result = run_test(new_test_ctx)
+                if test_result.exitcode != Exitcode.PASS.value:
+                    return test_result, next_exs
+                next_exs.append(post_ex)
+
+    return test_result, next_exs
+
+
+def run_target_contract(test_ctx, ex, addr) -> list[Exec]:
+
+    code = ex.code[addr]
+    contract_name = code.contract_name
+    filename = code.filename
+
+    if not contract_name:
+    #   if eq(addr, con_addr(FOUNDRY_TEST)):
+    #       debug_once(
+    #           f"{self.fun_info.sig}: couldn't find the contract name for: {hexify(selector)}"
+    #       )
+        return
+
+    contract_json = BuildOut().get_by_name(contract_name, filename)
+    abi = get_abi(contract_json)
+    method_identifiers = contract_json["methodIdentifiers"]
+
+    results = []
+
+    for fun_sig, fun_selector in method_identifiers.items():
+        fun_name = fun_sig.split("(")[0]
+        fun_info = FunctionInfo(fun_name, fun_sig, fun_selector)
+
+        new_test_ctx = FunctionContext(
+            args=test_ctx.args,
+            info=fun_info,
+            solver=mk_solver(test_ctx.args),
+            contract_ctx=test_ctx.contract_ctx,
+            setup_ex=ex,
+        )
+
+        # todo: mk_calldata with symbol name uniqueness
+        test_result = run_test(new_test_ctx)
+
+        results.extend(new_test_ctx.exec_cache.values())
+
+    return results
+
+
 def run_test(ctx: FunctionContext) -> TestResult:
     args = ctx.args
     fun_info = ctx.info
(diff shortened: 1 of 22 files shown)


src/halmos/__main__.py
```diff
@@ -394,7 +396,193 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+    exs = [setup_ex]
+
+    depth = 0
+    while True:
+        depth += 1
+        exs, funsigs = run_single_invariant_step(
+            ctx, exs, funsigs, test_results_map, depth
+        )
+
+        # todo: merge, simplify, or prioritize exs to mitigate path explosion
```
daejunpark, line 424: here we can add more optimization to reduce the number of paths

daejunpark: some ideas:
- filter out infeasible paths with a longer branching timeout
- prioritize paths based on uniqueness --- e.g., uncovered paths vs variations in byte sizes
- merge similar paths that have the same storage but slightly different path constraints

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
     return hevm_fail or any(is_global_fail_set(x) for x in context.subcalls())
 
 
+def get_state_id(ex: Exec) -> bytes:
+    """
+    Computes the state snapshot hash, incorporating constraints on state variables.
+
+    Assumes constraints on state variables have been precomputed by running Exec.path_slice() after completing a transaction.
+    Do not use this during transaction execution.
```
0xkarmacoma, line 416: ominous 😅

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+
+    # mutable context for invariant testing loop over multiple depths
+    inv_ctx = InvariantContext(
+        contract_ctx=ctx,
+        visited=visited,
+        test_results_map=test_results_map,
+    )
+
+    depth = 0
+    while True:
```
0xkarmacoma, line 469: I imagine there is something in `step_invariant_tests` that stops when the max inv-depth is reached, but wouldn't it make sense to have the condition explicitly at this outer loop?

```suggestion
    while depth <= args.inv_depth:
```

daejunpark: no because different inv-depth can be given for different invariant tests.

0xkarmacoma: ah neat, let's make a note of it then

```suggestion
    # inv_depth can be overridden by specific test functions, so not checked here
    while True:
```

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+    ctx = inv_ctx.contract_ctx
+    test_results_map = inv_ctx.test_results_map
+    visited = inv_ctx.visited
+
+    next_exs = []
+
+    for idx, pre_ex in enumerate(pre_exs):
+        progress_status.update(f"{depth=} {len(visited)=} {idx=} {len(pre_exs)=}")
+
+        for addr in pre_ex.code:
```
0xkarmacoma, line 532: we'll need to follow this up by adding some way to select the targets 

from the foundry docs: 

```
targetContract(address newTargetedContract_)

Adds a given address to the _targetedContracts array. This set of addresses is used for the target contracts. This array overwrites the set of contracts that was deployed during the setUp.
```

```
excludeContract(address newExcludedContract_)

Adds a given address to the _excludedContracts array. This set of contracts is explicitly excluded from the target contracts.
```

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+                # ignore and report if halmos-errored
+                if subcall.is_stuck():
+                    error(
+                        f"{depth=}: addr={hexify(addr)}: {subcall.get_stuck_reason()}"
+                    )
+                    continue
+
+                # ignore if reverted
+                if subcall.output.error:
+                    continue
```
0xkarmacoma, line 552: what if it reverted with an assertion failure?

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+                funsigs = [r.name for r in test_results if r.exitcode == PASS]
+
+                # print call trace if failed, to provide additional info for counterexamples
+                if any(r.exitcode != PASS for r in test_results):
+                    print("Path:")
+                    print(indent_text(hexify(post_ex.path)))
+
+                    print("\nTrace:")
+                    render_trace(post_ex.context)
+
```
0xkarmacoma, line 584: I noticed a minor annoyance around this:

```
Trace:
⠇ [0:00:00] solving queries: 0 / 1    CALL TStoreTest::setUp()
        CREATE 0xaaaa0002::<311 bytes of initcode>
```

notice how the status is interspersed with the trace output, annoying. Did some quick testing and it seems like we can work around this by first calling `progress_status.stop()` before printing the path and the trace. The problem is that we then lose the live status if there are longer running invariant tests, so then we need to call `progress_status.start()` again, like this:

```
                # print call trace if failed, to provide additional info for counterexamples
                if any(r.exitcode != PASS for r in test_results):
                    progress_status.stop()

                    print("Path:")
                    print(indent_text(hexify(post_ex.path)))

                    print("\nTrace:")
                    render_trace(post_ex.context)
                    
                    progress_status.start()
```

We can introduce a little helper context manager that makes this pattern easier:

```
class suspend:
    """Context manager for temporarily suspending and resuming progress status updates."""
    def __init__(self, progress_status):
        self.progress_status = progress_status

    def __enter__(self):
        self.progress_status.stop()

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.progress_status.start()
```

The code then looks like this:

```suggestion
                if any(r.exitcode != PASS for r in test_results):
                    with suspend(progress_status):
                        print("Path:")
                        print(indent_text(hexify(post_ex.path)))

                        print("\nTrace:")
                        render_trace(post_ex.context)
```

wdyt?

src/halmos/__main__.py
0xkarmacoma: is this expected?

```
    /// @custom:halmos --inv-depth 2
    function invariant_transient_storage() public {
        // note: transient storage is reset after each tx in the invariant tx sequence
        assertEq(c.tload(0), 0); // pass
    }

    /// @custom:halmos --inv-depth 4
    function invariant_transient_storage_long_running() public {
        assert(true);
    }
```

```
[PASS] invariant_transient_storage() (depth: 4, paths: 64)
[PASS] invariant_transient_storage_long_running() (depth: 4, paths: 64)
```

(the report shows them at the same depth, despite having a different config)

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+            # create a symbolic message value
+            msg_value = BitVec(
+                f"msg_value_{id_str(addr)}_{uid()}_{ex.new_symbol_id():>02}",
+                BitVecSort256,
+            )
+
+            # construct the transaction message
+            message = Message(
+                target=addr,
+                caller=ex.this(),
```
0xkarmacoma, line 662: I think we probably want a symbolic sender here -- maybe just avoiding address(0) and known contract addresses?

foundry behavior:

> Target Senders: The invariant test fuzzer picks values for msg.sender at random when performing fuzz campaigns to simulate multiple actors in a system by default. If desired, the set of senders can be customized in the setUp function.

src/halmos/__main__.py
```diff
@@ -761,21 +1021,73 @@ def run_contract(ctx: ContractContext) -> list[TestResult]:
...
     test_results = []
-    for funsig in ctx.funsigs:
+    for funsig in funsigs:
         selector = ctx.method_identifiers[funsig]
         fun_info = FunctionInfo(funsig.split("(")[0], funsig, selector)
         try:
             test_config = with_devdoc(args, funsig, ctx.contract_json)
+            # TODO: reuse solver across different functions
             solver = mk_solver(test_config)
             debug(f"{test_config.formatted_layers()}")
```
0xkarmacoma, line 1076: we probably need to remove this during an invariant run, or maybe have a separate `--debug-config` flag?

src/halmos/__main__.py
```diff
@@ -761,21 +1021,73 @@ def run_contract(ctx: ContractContext) -> list[TestResult]:
...
+    """
+    args = ctx.args
+
     test_results = []
-    for funsig in ctx.funsigs:
+    for funsig in funsigs:
         selector = ctx.method_identifiers[funsig]
         fun_info = FunctionInfo(funsig.split("(")[0], funsig, selector)
         try:
             test_config = with_devdoc(args, funsig, ctx.contract_json)
```
0xkarmacoma, line 1073: we need to move this out of `run_tests`, we spend a considerable amount of time parsing the config during an invariant test. Ok if I make a commit for it on your branch?

src/halmos/__main__.py
daejunpark: it's due to a bug in printing test results. currently displaying invariant test results is messy, and needs a proper refactoring to improve, which i'd like to handle in a separate pr.

src/halmos/config.py
```diff
@@ -274,6 +274,12 @@ class Config:
         action=ParseErrorCodes,
     )
 
+    inv_depth: int = arg(
```
0xkarmacoma, line 277: nit: I like full names better, can we call this `invariant_depth`?

src/halmos/config.py
```diff
@@ -274,6 +274,12 @@ class Config:
...
+        help="set depth for invariant testing",
```
0xkarmacoma, line 278:
```suggestion
        help="set depth for invariant testing (length of the stateful sequence of calls)",
```

src/halmos/__main__.py
daejunpark: it's for target contracts, and we don't expect assert cheatcodes there.

src/halmos/__main__.py
```diff
@@ -397,6 +408,279 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+                post_id = get_state_id(post_ex)
+                if post_id in visited:
+                    continue
+
+                # update visited set
+                # TODO: check path feasibility
+                visited.add(post_id)
+
+                # update call traces
+                post_ex.context = deepcopy(pre_ex.context)
```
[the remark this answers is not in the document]
daejunpark: we can improve it by removing setup trace, etc., but this particular code logic is needed to keep track of (and print the trace of) tx sequences.

src/halmos/__main__.py
daejunpark: i agree it's annoying. as i mentioned, displaying test results needs a proper refactoring, which would avoid this in the first place.

src/halmos/__main__.py
daejunpark: i agree. i prefer a separate flag for this, or having a multi-level debug flag for longer term.

src/halmos/__main__.py
daejunpark: agree. but i'd prefer to handle it in a separate pr, perhaps along with improving test result display. i'm afraid this pr might become too big.

src/halmos/config.py
daejunpark: done: https://github.com/a16z/halmos/pull/454/commits/5c6356b250564a5b2c15a1a132d7a8964187121c

src/halmos/__main__.py
daejunpark: symbolic sender and origin: https://github.com/a16z/halmos/pull/454/commits/de984d1897223db613b3a19f9ac1edef7c96bff9

src/halmos/__main__.py
```diff
@@ -397,6 +408,280 @@ def is_global_fail_set(context: CallContext) -> bool:
...
+        A list of invariant test function signatures remaining for the next depth.
+    """
+    ctx = inv_ctx.contract_ctx
+    test_results_map = inv_ctx.test_results_map
+    visited = inv_ctx.visited
+
+    next_exs = []
+
+    for idx, pre_ex in enumerate(pre_exs):
+        progress_status.update(f"{depth=} {len(visited)=} {idx=} {len(pre_exs)=}")
```
0xkarmacoma, line 531:
```suggestion
        progress_status.update(
            f"depth: {cyan(depth)} | "
            f"starting states: {cyan(len(pre_exs))} | "
            f"unique states: {cyan(len(visited))} | "
            f"frontier states: {cyan(len(next_exs))} | "
            f"completed paths: {cyan(idx)} "
        )
```

(requires importing `cyan`)

src/halmos/__main__.py
daejunpark: added in the todo list

src/halmos/__main__.py
daejunpark: added in the todo list

src/halmos/__main__.py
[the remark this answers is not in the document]
daejunpark: added in the todo list

0xkarmacoma (approved):
noted a improvements for later, but this is looking fantastic as-is! 🔥

src/halmos/__main__.py
daejunpark: added in the todo list

src/halmos/__main__.py
daejunpark: support for finer control of msg.sender was added in the todo list

src/halmos/__main__.py
daejunpark: added to the todo list

src/halmos/__main__.py
0xkarmacoma: would be extra cool to also display the last funsig attempted, so we would get a feel for where the exploration is headed. I can do that in a follow up!

Issue #393, opened 2024-10-14
support for Uniswap v2 testing workflow


0xkarmacoma:
**Describe the bug**

Different sha3_expr in in create2 init code lead to different inferred address vs actually deployed address.

In the example below, `uniswapV2Factory.createPair(address(token), address(weth))` results in the following sha3_expr:

    sha3_expr=f_sha3_680(Concat(372682917519380244141939632342652170012265661726725,
                  f_sha3_320(4184670640661422675847470105623367512667768574483259588612),
                  f_sha3_93088(...init code concrete value...)))

But in the [pairFor](https://github.com/Uniswap/v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol#L18) helper function, this is the expression used:

        pair = address(uint(keccak256(abi.encodePacked(
                hex'ff',
                factory,
                keccak256(abi.encodePacked(token0, token1)),
                hex'96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f' // init code hash
            ))));

(note the hardcoded initcode hash value)

On the halmos side, it results in the following sha3_expr:

    sha3_expr=f_sha3_680(Concat(372682917519380244141939632342652170012265661726725, 
                  f_sha3_320(4184670640661422675847470105623367512667768574483259588612), 
                  68258024698789349894765071221733254692395760896603232917352833510768193864799))

If we were to evaluate `f_sha3_93088(...init code concrete value...)` concretely, we would indeed get the same value:

```py
keccak(bytes.fromhex(hex(a)[2:])).hex()
'96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f'
```

Because of the sha3_expr mismatch, pairFor returns a different address (`bbbb0011`) than the address actually deployed (`bbbb000b`) and subsequent calls fail

**To Reproduce**

necessary steps from https://github.com/igorganich/damn-vulnerable-defi-halmos/blob/master/test/puppet-v2/PuppetV2_Halmos.t.sol#L86 (h/t @igorganich)

```solidity
    function setUp() public {
        // Deploy tokens to be traded
        token = new DamnValuableToken();
        weth = new WETH();

        // Deploy Uniswap V2 Factory and Router
        uniswapV2Factory = IUniswapV2Factory(/* deploy here */);
        uniswapV2Router = IUniswapV2Router02(/* deploy here */);

        // Create Uniswap pair
[...]

PR #462, opened 2025-02-16
fix: we do support empty bytes now

0xkarmacoma:


daejunpark:
the ci failure is orthogonal to the change in the test. but the change somehow triggers solc optimization, which replaces `keccak256("")` with its evaluation at compile time. this replacement doesn't use the push opcode, so [the existing mechanism](https://github.com/a16z/halmos/blob/5c5ca39a1ee943ad8c8dc2fe042bdea44413ed69/src/halmos/sevm.py#L3237) doesn't work.

this temporary workaround https://github.com/a16z/halmos/pull/462/commits/ae0cacb89feb090fc16752ed7463e1e1b6405437 fixes the ci failure, but it needs to be properly implemented. do not merge.

daejunpark:
it turns out that the empty keccak value is read from memory (probably because it was previously stored into memory using codecopy).  https://github.com/a16z/halmos/pull/462/commits/ca7c6e3157ba3ac1f7d76d0de1746d0e048637a9 moves the special value handling logic to the point where a value is pushed to the stack to cover various cases.

src/halmos/cheatcodes.py
```diff
@@ -407,20 +413,20 @@ def create_int(ex, arg, **kwargs):
...
 def create_bytes(ex, arg, **kwargs):
-    byte_size = int_of(
-        extract_bytes(arg, 4, 32), "symbolic byte size for halmos.createBytes()"
-    )
+    err = "symbolic byte size for halmos.createBytes()"
+    byte_size = int_of(extract_bytes(arg, 4, 32), err)
+    bits = byte_size * 8
     name = name_of(extract_string_argument(arg, 1))
-    symbolic_bytes = create_generic(ex, byte_size * 8, name, "bytes")
+    symbolic_bytes = create_any(ex, bits, name, "bytes") if bits else ByteVec()
```
daejunpark, line 429: an alternative approach is to add the empty byte logic inside create_any() rather than here.  though, it's unclear to me which is necessarily better.

if we decide to keep it here, then we also need to add it to create_string() below, which is basically the same as create_bytes() except for the error message.  indeed, it would be better to factor out the common logic to avoid code duplication.

daejunpark: https://github.com/a16z/halmos/pull/488

0xkarmacoma:
seems related to https://github.com/a16z/halmos/issues/393, but does not fix it

0xkarmacoma:
trying to wrap my head around this. It does feel like we're playing whack-a-mole since there could be other sources of concrete hashes (returndata, calldata, codecopy, ...). So we may end up with values on the stack that are either a concrete hash value, or its hash expression. In that sense, it would indeed be more consistent to eagerly lookup every value that goes on the stack in sha_inv and try to make it an expr, but at a cost. OTOH, I'm not completely sure I understand what issues we solve by eagerly converting everything to an expr. Some things that come to mind:
- storage slot decoding: it relies on being able to parse the expr, and so a concrete value is no good, especially for the solidity model
- deployment addresses: https://github.com/a16z/halmos/issues/393
- more?

Maybe an alternative would be that we don't need to eagerly do anything, and instead try to convert lazily to an expr at the point of use -- that could be both robust and cheap?

I say let's merge this PR as-is and do a more intentional keccak refactor in a follow-up

PR #470, opened 2025-03-05
perf: fast concrete stack using new bitvec API

0xkarmacoma:
Performance impact is going to be varied, but is significant for mostly concrete executions:

Before: 34859 ops/s
After: 595532 ops/s


pyproject.toml
```diff
@@ -24,7 +24,7 @@ requires-python = ">=3.11"
 dependencies = [
     "sortedcontainers>=2.4.0",
     "toml>=0.10.2",
-    "z3-solver==4.12.6.0",
+    "z3-solver==4.13.4.0",
```
0xkarmacoma, line 27: can't remember why we pinned to 4.12 but can undo this if you want

src/halmos/sevm.py
```diff
@@ -1894,192 +1925,69 @@ def mk_mod(self, ex: Exec, x: Any, y: Any) -> Any:
...
-                    bitsize = i2.bit_length() - 1
-                    return ZeroExt(w2.size() - bitsize, Extract(bitsize - 1, 0, w1))
-
-            return self.mk_mod(ex, w1, w2)
+            term = w1.mod(w2, abstraction=f_mod[w1.size])
+            if term.is_symbolic:
+                # (x % y) <= y
+                # not ULT, because y could be 0 and x % 0 = 0
+                ex.path.append(ULE(term.as_z3(), w2.as_z3()))
+            return term
```
0xkarmacoma, line 1952: not sure why yet but this seems to cause a problem in `check_Mod`, even though we do add `ULE(f_evm_bvurem_256(p_x_uint256_9834528_00, p_y_uint256_7a9fba7_00), p_y_uint256_7a9fba7_00)` to the path

this is the failing assert:

```
uint x_mod_y = unchecked_mod(x, y);
assert(x_mod_y <= y);
```

🤨

src/halmos/bitvec.py
```diff
@@ -0,0 +1,899 @@
...
+        return (
+            HalmosBool(Not(self._value))
+            if self._symbolic
+            else HalmosBool(not self._value)
+        )
+
+    def eq(self, other: "HalmosBool") -> "HalmosBool":
+        return HalmosBool(self._value == other._value)
+
+    def __and__(self, other: "HalmosBool") -> "HalmosBool":
```
daejunpark, line 213: why keeping `__and__` while not using `__add__` for BV?

daejunpark (commented):
skimmed through it, and left some random comments.

pyproject.toml
daejunpark: let's handle this separately, if possible.

src/halmos/sevm.py
```diff
@@ -3405,11 +3339,19 @@ def finalize(ex: Exec):
...
-                        codeslice: ByteVec = ex.pgm.slice(offset, size)
+                        # TODO: hide symbolic support behind a feature flag?
+                        # if the offset is symbolic, create a symbolic slice
+                        codeslice = (
+                            ex.pgm.slice(int(offset), size)
+                            if offset.is_concrete
+                            else ByteVec(
+                                BV(f"codeslice_uint{size * 8}_{uid()}", size=size * 8)
+                            )
+                        )
```
daejunpark, line 3354: is this adding a new feature or not? it would be better to have a new feature in a separate pr, if possible. actually, the usefulness of symbolic slice might need further discussion.

src/halmos/bitvec.py
```diff
@@ -498,6 +500,9 @@ def mul(
                 if is_power_of_two(rhs):
                     return self.lshl(HalmosBitVec(rhs.bit_length() - 1, size=size))
 
+                # simplify(x * val) returns val * x
+                return HalmosBitVec(rhs * lhs, size=size)
+
```
daejunpark, line 505: 👍

PR #472, opened 2025-03-10
feat: re-architect invariant testing + add dai invariant tests

daejunpark:
re-architected the invariant testing run to utilize the existing mechanism for parallel solving.

previously, the invariant testing run had its own pipeline, where the main loop computed the frontier states for each depth, and all invariant tests were executed together against each frontier state.

now, the invariant testing run is integrated into an extended version of the regular testing run, where each test can take multiple input states. for invariant testing, the input states consist of the frontier states (up to a given depth). these frontier states are computed once and cached to avoid recomputation across different tests.

also added dai invariant test examples.


daejunpark:
note for review: removing the previous logic (https://github.com/a16z/halmos/pull/472/commits/6d2b4aa2b24072834fcfe4f8dfe2b111c02b2afd) and adding a new logic makes the diff quite hard to parse. it would be easier to review each commit individually.

daejunpark:
another note: i encountered a test failure not related to this pr, and i had to fix it in this commit https://github.com/a16z/halmos/pull/472/commits/a683033957c4d2410e86d351b5b0e9b68ed9f49e   but i'm not sure why it didn't fail before.  anyway, could you double-check if the fix is correct?

src/halmos/__main__.py
```diff
@@ -593,58 +517,32 @@ def step_invariant_tests(
                 # TODO: check path feasibility
                 visited.add(post_id)
 
-                # transfer the setup_ex message to the post_ex message
-                # (so that invariant tests run against the test contract)
-                post_ex.context.message = pre_ex.context.message
```
daejunpark, line 598: note: this is no longer needed because now we explicitly use TEST_TARGET, TEST_CALLER, TEST_ORIGIN when creating a message for test contracts.

src/halmos/__main__.py
```diff
@@ -238,17 +239,26 @@ def mk_solver(args: HalmosConfig, logic="QF_AUFBV", ctx=None):
     )
 
 
+# NOTE: Do NOT remove the `con_addr()` wrapper for TEST_TARGET.
+# The return type should be BitVecSort(160) as it is used as a key for ex.code.
+# The keys of ex.code are compared using structural equality with other BitVecRef addresses.
+TEST_TARGET = con_addr(FOUNDRY_TEST)
+TEST_CALLER = FOUNDRY_CALLER
+TEST_ORIGIN = FOUNDRY_ORIGIN
```
0xkarmacoma, line 247: why introduce aliases instead of just using the `FOUNDRY_*` constants?

daejunpark: ah just wanted to use the same prefix for all three.  we can also change FOUNDRY_TEST to be `con_addr(0x...)`, and use the `FOUNDRY_*` prefix.  do you prefer that?

0xkarmacoma: yeah that works! 👍

src/halmos/__main__.py
```diff
@@ -730,7 +626,46 @@ def run_target_contract(ctx: ContractContext, ex: Exec, addr: Address) -> list[E
...
+    """
+    Retrieves the frontier states at a given depth.
+
+    If the frontier states have already been computed, the cached results are returned.
+    Otherwise, the generator from compute_frontier() is returned.
+    """
+    frontier_states = ctx.frontier_states
+
+    if depth in frontier_states:
+        return frontier_states[depth]
```
0xkarmacoma, line 640:
```suggestion
    if (frontier := frontier_states.get(depth)) is not None:
        return frontier
```

src/halmos/__main__.py
```diff
@@ -730,7 +626,46 @@ def run_target_contract(ctx: ContractContext, ex: Exec, addr: Address) -> list[E
...
+    A frontier state is the result of executing a sequence of arbitrary txs starting from the initial setup state.
+    These states are grouped by their tx depth (i.e., the number of txs in a sequence) and cached in ContractContext.frontier_states to avoid re-computation for other tests.
+
+    The max tx depth to consider is specified in FunctionContext, which is given by --invariant-depth for invariant tests, and set to 0 for regular tests.
+
+    For regular tests (where the max tx depth is 0), this function amounts to executing the given test against only the initial setup state.
+    """
+    for depth in range(ctx.max_call_depth + 1):
+        for ex in get_frontier(ctx.contract_ctx, depth):
+            reset(ctx.solver)
```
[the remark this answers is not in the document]
daejunpark: i thought about it too, but i wasn't sure if the early-exit mechanism would still work in that case. wdyt?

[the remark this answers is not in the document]
0xkarmacoma: I think these are unrelated:
- `ctx.solver` is the branching solver
- early exit relies on assertion solvers in external processes

src/halmos/__main__.py
0xkarmacoma: (it doesn't really matter in this case, but in general I prefer the pattern of doing a single lookup vs 2 lookups in a dict)

src/halmos/__main__.py
```diff
@@ -730,7 +626,46 @@ def run_target_contract(ctx: ContractContext, ex: Exec, addr: Address) -> list[E
...
 
-    return results
+
+def get_frontier(ctx: ContractContext, depth: int) -> Iterable[Exec]:
+    """
+    Retrieves the frontier states at a given depth.
+
+    If the frontier states have already been computed, the cached results are returned.
+    Otherwise, the generator from compute_frontier() is returned.
+    """
```
0xkarmacoma, line 636: makes sense, I would add a note that this is not thread safe, it is meant to be only called from a single thread. Otherwise we could potentially get things like an incomplete `frontier` (while `compute_frontier` is still running)

src/halmos/__main__.py
```diff
@@ -422,120 +431,34 @@ def get_state_id(ex: Exec) -> bytes:
     return snapshot_state(ex, include_path=True).unwrap()
 
 
-def run_invariant_tests(
-    ctx: ContractContext, pre_ex: Exec, funsigs: list[str]
-) -> list[TestResult]:
+def compute_frontier(ctx: ContractContext, depth: int) -> Iterator[Exec]:
```
0xkarmacoma, line 434: I would name this `_compute_frontier` to signal that this is an "internal" kind of function and make a note that this is only to be called by `get_frontier`.

0xkarmacoma (approved):

src/halmos/__main__.py
daejunpark: https://github.com/a16z/halmos/pull/472/commits/b338f2a1a5c87b79d51a13d443db54979dc49955

src/halmos/__main__.py
daejunpark: https://github.com/a16z/halmos/pull/472/commits/7b9503493e36795034fd37cfce6423dfbd9bd7ba

src/halmos/__main__.py
daejunpark: https://github.com/a16z/halmos/pull/472/commits/7b9503493e36795034fd37cfce6423dfbd9bd7ba

src/halmos/__main__.py
daejunpark: https://github.com/a16z/halmos/pull/472/commits/7b9503493e36795034fd37cfce6423dfbd9bd7ba

src/halmos/__main__.py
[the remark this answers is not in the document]
daejunpark: https://github.com/a16z/halmos/pull/472/commits/8763dc9bbec56946860931fd0878be60cd58fc3b

src/halmos/__main__.py
```diff
@@ -422,128 +415,140 @@ def get_state_id(ex: Exec) -> bytes:
...
-    ctx = inv_ctx.contract_ctx
-    test_results_map = inv_ctx.test_results_map
-    visited = inv_ctx.visited
-    panic_error_codes = ctx.args.panic_error_codes
+    # frontier states at the previous depth, which will be used as input for computing new frontier states at the current depth
+    curr_exs = ctx.frontier_states[depth - 1]
 
-    next_exs = []
+    # the cache for the new frontier states
+    next_exs = ctx.frontier_states[depth]
```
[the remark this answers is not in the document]
daejunpark: https://github.com/a16z/halmos/pull/472/commits/3c675541a3b7225788c8f1f1c9a3e12b42a081e4

PR #473, opened 2025-03-11
optimizing the interpreter loop based on instruction profiles

0xkarmacoma:
before: ~600k ops/s
after: ~1.4M ops/s

(note: builds on top of perf/concrete-stack)


0xkarmacoma:
(looks like ci is not running because it does not target the main branch, will figure this out once the base branch has been merged)

daejunpark (approved):

src/halmos/sevm.py
```diff
@@ -709,7 +708,7 @@ def popi(self) -> BV:
         """The stack can contain BitVecs or Bools -- this function converts Bools to BitVecs"""
 
         val = self.pop()
-        return val if type(val) is BV else BV(val, size=256)
+        return val.as_bv(size=256) if type(val) is Bool else val
```
daejunpark, line 711: how can we make sure that val is either bool or bv here?  not specific to this, but overall, all the various optimizations here and there make it much harder to reason about the type consistency manually. perhaps we would need to consider incorporating actual type checking (not just type hinting), like mypy?

0xkarmacoma: yes, mypy/pyright would be amazing, but that's probably going to be a big lift to get there

there is an assert in State.push:

```
    def push(self, v: Bool | BV) -> None:
        type_v = type(v)
        assert type_v is BV and v.size == 256 or type_v is Bool
        self.stack.append(v)
```

the assert is on by default and a little bit costly at runtime, but we know for sure that State.stack only contains Bools and word-size BVs

0xkarmacoma:
> makes sense overall, but i'm not sure about inlining jump() and jumpi(). is the performance improvement big enough to justify compromising code readability?

I agree we need to be careful, I didn't feel like I was compromising code readability too much. When you scroll through the opcode dispatcher now, it's still a bit of a monster, but it does not completely unreasonable to me.

PR #477, opened 2025-03-14
feat: report assertion failures in contract code during invariant testing (probes)

0xkarmacoma:
Example:

```
$ halmos --root tests/regression --function invariant_probes_found

[⠒] Compiling...
No files changed, compilation skipped

Running 1 tests for test/InvariantProbes.t.sol:InvariantTest
Assertion failure detected in WithProbes.probe0()
Sequence:
    (empty)

Trace:
    CALL 0xaaaa0002::probe0() (value: msg_value_0x00000000000000000000000000000000aaaa0002_f8942ae_06)
    ↩ REVERT 0x4e487b710000000000000000000000000000000000000000000000000000000000000001 (error: Revert())

Assertion failure detected in WithProbes.probe1()
Sequence:
    CALL 0xaaaa0002::inc() (value: msg_value_0x00000000000000000000000000000000aaaa0002_8645bbe_03)
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000000
        SSTORE @0 ← 0x0000000000000000000000000000000000000000000000000000000000000001
    ↩ STOP 0x

Trace:
    CALL 0xaaaa0002::probe1() (value: msg_value_0x00000000000000000000000000000000aaaa0002_010eba8_12)
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000001
    ↩ REVERT 0x4e487b710000000000000000000000000000000000000000000000000000000000000001 (error: Revert())

Assertion failure detected in WithProbes.probe2()
Sequence:
    CALL 0xaaaa0002::inc() (value: msg_value_0x00000000000000000000000000000000aaaa0002_8645bbe_03)
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000000
        SSTORE @0 ← 0x0000000000000000000000000000000000000000000000000000000000000001
    ↩ STOP 0x
    CALL 0xaaaa0002::inc() (value: msg_value_0x00000000000000000000000000000000aaaa0002_4dafc22_06)
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000001
        SSTORE @0 ← 0x0000000000000000000000000000000000000000000000000000000000000002
    ↩ STOP 0x
    CALL 0xaaaa0002::probe1() (value: msg_value_0x00000000000000000000000000000000aaaa0002_0657858_15)
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000002
        SLOAD  @0 → 0x0000000000000000000000000000000000000000000000000000000000000002
        SLOAD  @1 → 0x0000000000000000000000000000000000000000000000000000000000000000
        SSTORE @1 ← 0x0000000000000000000000000000000000000000000000000000000000000001
    ↩ STOP 0x

Trace:
    CALL 0xaaaa0002::probe2() (value: msg_value_0x00000000000000000000000000000000aaaa0002_f4873bc_27)
        SLOAD  @1 → 0x0000000000000000000000000000000000000000000000000000000000000001
    ↩ REVERT 0x4e487b710000000000000000000000000000000000000000000000000000000000000001 (error: Revert())

[PASS] invariant_probes_found() (depth: 4, paths: 19)
Symbolic test result: 1 passed; 0 failed; time: 0.62s
```

note:
- readable names (e.g. `Assertion failure detected in WithProbes.probe1()`)
- each probe is only reported once, but is still used in state exploration
- separation of call sequence (how we got there) and the trace of the last call
- this currently does not fail the test, should it?


src/halmos/__main__.py
```diff
@@ -571,8 +592,8 @@ def step_invariant_tests(
                 visited.add(post_id)
 
                 # update call traces
-                post_ex.context = deepcopy(pre_ex.context)
-                post_ex.context.trace.append(subcall)
+                post_ex.context = copy(pre_ex.context)
```
daejunpark, line 595: isn't the trace of this shallow-copied context supposed to be updated when pre_ex.trace is extended later? or, do we not need to worry because pre_ex.trace will never be extended?

daejunpark (approved):
i have a question about shallow vs deep copy of context, but other than that, looks good to me. thank you!

src/halmos/__main__.py
0xkarmacoma: well tbh deepcopy didn't feel right, but it's also not entirely clear to me why we should copy at all. Semantically, that doesn't make a lot of sense to me. The trace shouldn't be extended, that would be weird to me (but the call sequence should be!)

Can you check again with d329b7a? No copy at all, turns out the only part we need from the context is the immutable Message from setup_ex.

0xkarmacoma: (eventually I would like to have a separate NetworkState field in Exec, so that we can explicitly pluck out the state from an exec and discard intermediate values like the EVMState and CallContext)

daejunpark: > Can you check again with https://github.com/a16z/halmos/commit/d329b7a25d814c7afdca9a628132f4ab09fe9211? No copy at all, turns out the only part we need from the context is the immutable Message from setup_ex.

looks good to me, thank you!

PR #487, opened 2025-03-27
add support for targetSender(), targetContract(), and targetSelector()

daejunpark:
added support for invariant testing target cheatcodes.

also improved test outputs.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -23,11 +23,11 @@ from importlib import metadata
 import rich
 from z3 import (
     BitVec,
+    BoolRef,
+    Or,
     ZeroExt,
-    eq,
     set_option,
     unsat,
-    Or,
 )
 
 import halmos.traces
@@ -97,21 +97,22 @@ from .solve import (
 )
 from .traces import render_trace, rendered_call_sequence, rendered_trace
 from .utils import (
-    con_addr,
-    extract_bytes,
-    int_of,
     EVM,
     Address,
     BitVecSort256,
     NamedTimer,
+    Word,
     address,
     color_error,
     con,
+    con_addr,
     create_solver,
     cyan,
+    extract_bytes,
     green,
     hexify,
     indent_text,
+    int_of,
     red,
     uid,
     unbox_int,
@@ -292,6 +293,7 @@ def deploy_test(ctx: FunctionContext, sevm: SEVM) -> Exec:
         )
 
     deployed_bytecode = Contract(returndata)
+    ex.try_resolve_contract_info(deployed_bytecode)
     ex.set_code(this, deployed_bytecode)
     ex.pgm = deployed_bytecode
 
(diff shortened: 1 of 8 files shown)


src/halmos/__main__.py
```diff
@@ -415,6 +421,52 @@ def get_state_id(ex: Exec) -> bytes:
...
+        path.extend_path(ex.path)
+
+        # prepare calldata and dynamic parameters
+        calldata, dyn_params = mk_calldata(
+            abi, fun_info, args, new_symbol_id=ex.new_symbol_id
+        )
+        path.process_dyn_params(dyn_params)
+
+        # add (optional) constraints on msg_sender
+        if msg_sender_cond:
```
0xkarmacoma, line 449: doesn't this give an error if `msg_sender_cond` is an actual BoolRef? does it mean this is never called?

```suggestion
        if msg_sender_cond is not None:
```

0xkarmacoma: indeed adding a call to `targetSender(x)` in one of the invariant tests results in this error:

```
ERROR    run_target_contract 2863267842 inc(uint256): Z3Exception: Symbolic expressions cannot be cast to concrete Boolean values.                                                                                                                                                                                      
```

src/halmos/__main__.py
```diff
@@ -459,57 +513,47 @@ def run_target_contract(
...
-                f"msg_sender_{id_str(addr)}_{uid()}_{ex.new_symbol_id():>02}"
+                f"halmos_msg_sender_{id_str(addr)}_{uid()}_{ex.new_symbol_id():>02}"
+            )
+
+            # restrict msg.sender to the specified target senders
+            msg_sender_cond = (
+                Or([msg_sender == target_sender for target_sender in target_senders])
+                if target_senders
+                else None
             )
```
0xkarmacoma, line 531: we can avoid creating an Or expression with target_senders is a single element:

```suggestion
            msg_sender_cond = (
                msg_sender == target_senders[0]
                if len(target_senders) == 1
                else Or([msg_sender == sender for sender in target_senders])
                if target_senders
                else None
            )
```

src/halmos/__main__.py
daejunpark: oh good catch!

daejunpark: https://github.com/a16z/halmos/pull/487/commits/1cd0c3a27260782d19c38ff84c320222484c8142

src/halmos/__main__.py
daejunpark: https://github.com/a16z/halmos/pull/487/commits/ca4f6b3a0da4cc20adcfa7774fdc6f7c809332c7

src/halmos/__main__.py
```diff
@@ -986,6 +1031,218 @@ def reset(solver):
     solver.reset()
 
 
+def setup_invariant_test_context(ctx: ContractContext, setup_ex: Exec):
+    # skip if forge-std/Test.sol is not imported
+    if "targetSenders()" not in ctx.abi:
```
0xkarmacoma, line 1036: that's a clever check!

0xkarmacoma (approved):
looks great! TBH I haven't completely internalized the rules and precedence around target contracts and selectors, it may be useful to write some guidelines around what we think is the right way to use this

daejunpark:
> I haven't completely internalized the rules and precedence around target contracts and selectors, it may be useful to write some guidelines around what we think is the right way to use this

right, but actually it matches foundry's behavior, so i believe users are already familiar with it.

PR #489, opened 2025-03-28
experimental: add flamegraph generation support

0xkarmacoma:
This supports generating 2 kinds of flamegraphs, enabled with the `--flamegraph` option
- `exec-flamegraph.svg`: renders all paths from all tests 
- `call-flamegraph.svg`: renders call sequences and failed invariant checks

Examples: 

- execution flamegraph from tests/regression:
image

- call flamegraph for `VatTest`:
image
image

- call flamegraph for `HardMaze`: 
image

**UPDATE:** Some additional goodies below

## Differential Flamegraphs

```
# run 1, generates call.stacks and call-flamegraph.svg
halmos --flamegraph
mv call.stacks old.stacks

# run 2
halmos --flamegraph
mv call.stacks new.stacks

difffolded.pl old.stacks new.stacks | flamegraph.pl > diff.svg
```

Shows increases in red and decreases in blue between the first and the second run. Can potentially translate to increased/decreased coverage. For instance if we added a public function that can be reached during invariant testing, we expect to see it in red in the flamegraph diff (see https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html).

image

## Live Reloads

Because invariant runs can run a long time and may be interrupted, we don't want to wait until the end of the run to render the flamegraphs. So in [2201c98](https://github.com/a16z/halmos/pull/489/commits/2201c984611ea2f2dc0fccd06a2c60d43fb10178) we added on-the-fly generation: we periodically write the folded call stacks to `call.stacks` and invoke `flamegraph.pl` on it to render `call-flamegraph.svg`. So it is possible to live reload `call-flamegraph.svg` to visualize the execution as it happens:

```
# in one shell:
halmos --function invariant_XYZ --flamegraph

# in another shell:
npx live-server call-flamegraph.svg
```

This will open a browser window that will auto-refresh whenever a new version of `call-flamegraph.svg` is generated.

flamegraphs

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -50,6 +50,7 @@ from .constants import (
     VERBOSITY_TRACE_SETUP,
 )
 from .exceptions import FailCheatcode, HalmosException
+from .flamegraphs import flamegraph
 from .logs import (
     COUNTEREXAMPLE_INVALID,
     COUNTEREXAMPLE_UNKNOWN,
@@ -276,6 +277,9 @@ def deploy_test(ctx: FunctionContext, sevm: SEVM) -> Exec:
 
     [ex] = exs
 
+    if ctx.args.coverage_flamegraph:
+        flamegraph.add(ex.context)
+
     if ctx.args.verbose >= VERBOSITY_TRACE_CONSTRUCTOR:
         print("Constructor trace:")
         render_trace(ex.context)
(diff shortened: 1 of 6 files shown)


igorganich:
Will it cover only completed paths or even non-finished paths, that was stopped, for example, by using "vm.assume(false)"?

0xkarmacoma:
> Will it cover only completed paths or even non-finished paths, that was stopped, for example, by using "vm.assume(false)"?
> 
> I think it may be very informative to have info even for unfinished paths since we can use optimization via snapshotState and some paths will be unfinished

no unfinished (or rather infeasible) paths are not represented in the output because they are not exposed by sevm. I agree that it seems inconsistent, because you may see a `console.log` statement in the output, but it would not show up in the flamegraph. Need to think a little bit about how to handle this.

daejunpark (approved):

Issue #469, opened 2025-03-01
Make the usage of non-z3 solvers easier


montyly:
**Is your feature request related to a problem? Please describe.**
Right now using non-z3 solver is a bit tedious.

For example, if I am not mistaken:
- For yices, I need to use `-solver-command yices-smt2`
- For bitwuzla, I need to use `--solver-command "bitwuzla --produce-models"`

Etc. 

**Describe the solution you'd like**

I think having a cli flag `--solver NAME`, with the most common solvers will help (including jsi). 

(`-solver-command` will still be useful if someone wants to do a non-standard usage)

0xkarmacoma:
that's a great idea! can you imagine if we also offered to download the binary for you if you didn't have the solver installed? 👀

0xkarmacoma:
this is the approach used in hevm:

https://github.com/ethereum/hevm/blob/main/src/EVM/Solvers.hs#L335-L355

0xkarmacoma:
Compiling some of the args in the wild

```
# yices
--solver-command "yices-smt2 --smt2-model-format --bvconst-in-decimal"

# bitwuzla (but also cvc4 and cvc5)
--solver-command "bitwuzla --produce-models"

# boolector
--solver-command "boolector --model-gen --output-number-format=hex"

# z3
z3 --model

# jsi
--solver-command "jsi --model"
```

(see https://github.com/a16z/jsi/blob/main/src/jsi/config/solvers.json)

0xkarmacoma:
Sub-issue: #492

0xkarmacoma:
Fixed in [feat: add --solver <name> command, download on demand #491](https://github.com/a16z/halmos/pull/491)

PR #491, opened 2025-04-02
feat: add --solver <name> command, download on demand

0xkarmacoma:
implement the suggestion in https://github.com/a16z/halmos/issues/469, with just yices for now

- `--solver` defaults to z3, and `--solver-command` defaults to empty
- `--solver-command` is still available, it can be overriden anywhere and takes precedence over `--solver`
- when `--solver-command` is empty, it is resolved dynamically based on the selected solver name. We look for it in the binary cache (`~/.halmos/solvers`), and if it's not there we download it from a predefined releases URL, check its integrity against a known checksum and extract the binary in the halmos cache
- `--solver` can be overriden at the function level

Example flow:

```
# on first run:
$ halmos --debug --solver yices 
...
DEBUG    Solver binary 'yices' not found in cache or PATH.                                                                                                                                                                                                                                                              
INFO     Downloading from https://github.com/SRI-CSL/yices2/releases/download/Yices-2.6.5/yices-2.6.5-x86_64-apple-darwin21.6.0-static-gmp.tar.gz...                                                                                                                                                                    
INFO     Verifying sha256 hash for /var/folders/p9/s6g8495n6dqdjq3nn8hkdn840000gn/T/tmpjvb6319n/yices-2.6.5-x86_64-apple-darwin21.6.0-static-gmp.tar.gz...                                                                                                                                                              
INFO     Extracting yices-2.6.5/bin/yices-smt2 from /var/folders/p9/s6g8495n6dqdjq3nn8hkdn840000gn/T/tmpjvb6319n/yices-2.6.5-x86_64-apple-darwin21.6.0-static-gmp.tar.gz...                                                                                                                                             
INFO     Wrote /Users/karma/.halmos/solvers/yices (2.5MB)                                                                                                                                                                                                                                                               
DEBUG    Solver command for 'yices': /Users/karma/.halmos/solvers/yices --smt2-model-format

# on subsequent runs
$ halmos --debug --solver yices 
...
DEBUG    Found yices binary in cache: /Users/karma/.halmos/solvers/yices                                                                                                                                                                                                                                                
DEBUG    Solver command for 'yices': /Users/karma/.halmos/solvers/yices --smt2-model-format

# we look for z3 in the venv root, same as before
$ halmos --debug --solver z3
...
DEBUG    Found z3 binary in venv: /Users/karma/projects/halmos/.venv/bin/z3                                                                                                                                                                                                                                             
DEBUG    Solver command for 'z3': /Users/karma/projects/halmos/.venv/bin/z3  
```

Note that jsi doesn't follow the same one binary model, so for now it still has to be installed separately and invoked with `--solver-command`.

TODO:
- [ ] bitwuzla
- [ ] cvc5
- [x] verify windows support

### DIFF
--- a/src/halmos/solvers.py
+++ b/src/halmos/solvers.py
@@ -207,8 +207,9 @@ def install_solver(solver_info: SolverInfo) -> Path:
         machine_tuple = get_platform_arch()
         download_info = solver_info.downloads.get(machine_tuple)
         if not download_info:
-            raise ValueError(
-                f"No download URL configured for solver '{solver_info.name}'"
+            solver = solver_info.name
+            raise RuntimeError(
+                f"No download URL configured {solver=}, {machine_tuple=}"
             )
 
         # Download the archive
--- a/src/halmos/solvers.py
+++ b/src/halmos/solvers.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 import requests
 
-from halmos.logs import debug, error, info, warn
+from halmos.logs import debug, error, info
 from halmos.utils import format_size
 
 # Define the cache directory for solvers
@@ -40,13 +40,17 @@ class DownloadInfo:
 
 @dataclass(frozen=True, eq=False, order=False, slots=True)
 class SolverInfo:
+    # descriptive name, e.g., "yices" or "bitwuzla-abstraction"
     name: str
 
-    # maps (system, machine) tuples to download URLs
-    downloads: dict[MachineInfo, DownloadInfo]
+    # name of the executable, e.g., "yices-smt2" or "bitwuzla"
+    binary_name: str
 
     # options/arguments/flags needed for this solver
     arguments: list[str]
+
+    # maps (system, machine) tuples to download URLs
+    downloads: dict[MachineInfo, DownloadInfo]
 
 
 macos_intel = MachineInfo(system="Darwin", machine="x86_64")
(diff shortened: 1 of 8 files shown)


0xkarmacoma:
bitwuzla is tricky to support like this because they don't provide static release binaries, so we can't automatically download it, but we could provide the default arguments if the user has bitwuzla already installed

src/halmos/solvers.py
```diff
@@ -0,0 +1,400 @@
...
+import requests
+
+from halmos.logs import debug, error, info
+from halmos.utils import format_size
+
+# Define the cache directory for solvers
+SOLVER_CACHE_DIR = Path.home() / ".halmos" / "solvers"
+SOLVER_CACHE_DIR.mkdir(parents=True, exist_ok=True)
+
+YICES_BASE_URL = "https://github.com/SRI-CSL/yices2/releases/download/Yices-2.6.5"
```
daejunpark, line 21: we're using 2.6.4 because of the performance degradation in the latest version:
https://github.com/a16z/halmos/blob/main/packages/solvers/Dockerfile#L33

daejunpark (approved):
looks good to me. the only request is to change the yices version to 2.6.4.

daejunpark:
https://github.com/a16z/halmos/pull/491/commits/35c43e7f197cbba8fc46b3b7a8a7888fe7557a9e looks great, thanks!

PR #493, opened 2025-04-03
fix: self-transfer value

daejunpark:
the balance update logic was incorrect in the self-transfer case. this bug was introduced during the large refactoring in #470, but it wasn't caught by the existing tests. specifically, SendTest was supposed to catch this bug, but the test had been written under the assumption that initial setup balances were symbolic, which became invalid after we modified the initial setup balance to be zero.

this pr fixed both the balance update logic and the test.

### DIFF
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -2502,7 +2502,6 @@ class SEVM:
             return
 
         caller_balance: BitVecRef = ex.balance_of(caller)
-        to_balance: BitVecRef = ex.balance_of(to)
 
         # assume balance is enough; otherwise ignore this path
         # note: evm requires enough balance even for self-transfer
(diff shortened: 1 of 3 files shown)


daejunpark:
let me merge this quickly, as it fixes a soundness bug. especially, the [pectra formal verification](https://github.com/daejunpark/sys-asm-halmos) is currently broken due to it.

src/halmos/sevm.py
```diff
@@ -2517,7 +2516,8 @@ def transfer_value(
             value = If(condition, value, Z3_ZERO)
 
         ex.balance_update(caller, BV(caller_balance).sub(value))
-        ex.balance_update(to, BV(to_balance).add(value))
+        # NOTE: ex.balance_of(to) must be called **after** updating the caller's balance above, to correctly handle the self-transfer case
+        ex.balance_update(to, BV(ex.balance_of(to)).add(value))
```
0xkarmacoma, line 2520: good catch 😱

PR #484, opened 2025-03-20
fix: failed transfers revert

0xkarmacoma:
fixes https://github.com/a16z/halmos/issues/480

PR #494, opened 2025-04-03
test: add pectra external test

daejunpark:
the [pectra formal verification ci](https://github.com/daejunpark/sys-asm-halmos/actions) broke after #470 and #484. to prevent further regressions, the pectra test has been added to the external test ci. to reduce ci overhead, the loop bound is set to 2 in our ci (vs 16 in the original pectra fv ci).


0xkarmacoma (approved):
yes! more tests! 😁

PR #497, opened 2025-04-04
perf: reduce redundant simplification

daejunpark:


0xkarmacoma (approved):
wow I love this change! absolutely genius 🤩 comes from an absolutely brilliant mind, no doubt

daejunpark:
> wow I love this change! absolutely genius 🤩 comes from an absolutely brilliant mind, no doubt

which is you :)

PR #498, opened 2025-04-04
Fix/prompt status clash + cvc5 and bitwuzla configs

0xkarmacoma:
The problem is that if we display an input/prompt while there is a live status, the status updates will clobber the prompt and we'll be stuck. 

My solution is to introduce a central `ui` class in the `halmos.ui` module that holds the singleton status. As long as status updates and prompts go through this class, things work as expected.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -21,7 +21,6 @@ from datetime import timedelta
 from enum import Enum
 from importlib import metadata
 
-import rich
 from z3 import (
     BitVec,
     BoolRef,
@@ -61,7 +60,6 @@ from .logs import (
     error,
     logger,
     logger_unique,
-    progress_status,
     warn,
     warn_code,
 )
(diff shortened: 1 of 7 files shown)


src/halmos/__main__.py
```diff
@@ -1449,10 +1447,6 @@ def _main(_args=None) -> MainResult:
     timer = NamedTimer("total")
     timer.create_subtimer("build")
 
-    # clear any remaining live display before starting a new instance
-    rich.get_console().clear_live()
-    progress_status.start()
```
0xkarmacoma, line 1454: moving the status start to after the forge build command, which itself has a live status and they seemed to clash

src/halmos/solvers.py
```diff
@@ -307,8 +368,12 @@ def install_solver(solver: SolverInfo) -> Path:
         if install_path.exists():
             raise RuntimeError(f"File already exists: {install_path}")
 
+        ui.print(
+            f"Solver saved to [bold]{install_path}[/bold] ([cyan]{format_size(len(content))}[/cyan])",
+            highlight=False,
+            markup=True,
+        )
```
0xkarmacoma, line 375: experimenting with some richer output, looks like this

image

daejunpark (approved):

0xkarmacoma:
This is a follow up on https://github.com/a16z/halmos/pull/491

PR #501, opened 2025-04-07
deal with large vm.deal amounts

### LINKED ISSUE #338: WARNING:halmos:path.append(false) in Damn Vulnerable Defi - NaiveReceiver.t.sol
**Describe the bug**

As reported by https://t.me/c/1815219512/2277 and easily reproducible. We get the following warnings during the execution of `setUp()`:

```
WARNING:halmos:path.append(false)
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:halmos:path.append(false)
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
```

With some extra debug info:

```
Running 1 tests for test/naive-receiver/NaiveReceiver.t.sol:NaiveReceiverChallenge
Constructor trace:
    CREATE NaiveReceiverChallenge::<0 bytes of initcode>
    ↩ RETURN <20027 bytes of code>

  File "halmos/.venv/bin/halmos", line 8, in <module>
    sys.exit(main())
  File "halmos/src/halmos/__main__.py", line 1677, in main
    return _main().exitcode
  File "halmos/src/halmos/__main__.py", line 1641, in _main
    test_results = run_method(run_args)
  File "halmos/src/halmos/__main__.py", line 1019, in run_sequential
    setup_ex = setup(
  File "halmos/src/halmos/__main__.py", line 520, in setup
    for idx, setup_ex in enumerate(setup_exs_all):
  File "halmos/src/halmos/sevm.py", line 2415, in run
    self.create(ex, opcode, stack, step_id)
  File "halmos/src/halmos/sevm.py", line 1913, in create
    self.transfer_value(ex, caller, new_addr, value)
  File "halmos/src/halmos/sevm.py", line 1528, in transfer_value
    balance_cond = simplify(UGE(ex.balance_of(caller), value))
  File "halmos/src/halmos/sevm.py", line 817, in balance_of
    self.path.append(ULT(value, con(2**96)))
  File "halmos/src/halmos/sevm.py", line 598, in append
    import traceback; traceback.print_stack()
WARNING:halmos:path.append(false) orig_cond=ULT(340282366920938463463374607431768211456, 79228162514264337593543950336)
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
  File "halmos/.venv/bin/halmos", line 8, in <module>
    sys.exit(main())
  File "halmos/src/halmos/__main__.py", line 1677, in main
    return _main().exitcode
  File "halmos/src/halmos/__main__.py", line 1641, in _main
    test_results = run_method(run_args)
  File "halmos/src/halmos/__main__.py", line 1019, in run_sequential
    setup_ex = setup(
  File "halmos/src/halmos/__main__.py", line 520, in setup
    for idx, setup_ex in enumerate(setup_exs_all):
  File "halmos/src/halmos/sevm.py", line 2415, in run
    self.create(ex, opcode, stack, step_id)
  File "halmos/src/halmos/sevm.py", line 1913, in create
    self.transfer_value(ex, caller, new_addr, value)
  File "halmos/src/halmos/sevm.py", line 1537, in transfer_value
    ex.balance_update(caller, self.arith(ex, EVM.SUB, ex.balance_of(caller), value))
  File "halmos/src/halmos/sevm.py", line 817, in balance_of
    self.path.append(ULT(value, con(2**96)))
  File "halmos/src/halmos/sevm.py", line 598, in append
    import traceback; traceback.print_stack()
WARNING:halmos:path.append(false) orig_cond=ULT(340282366920938463463374607431768211456, 79228162514264337593543950336)
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
[console.log] msg.sender 0x00000000000000000000000000000000000000000000000000000000de4107e4
[console.log] trustedForwarder 0x00000000000000000000000000000000000000000000000000000000aaaa0003
Error: setUp() failed: HalmosException: No successful path found in setUp()
  File "halmos/src/halmos/__main__.py", line 1019, in run_sequential
    setup_ex = setup(
               ^^^^^^
  File "halmos/src/halmos/__main__.py", line 560, in setup
    raise HalmosException(f"No successful path found in {setup_sig}")
halmos.exceptions.HalmosException: No successful path found in setUp()
Symbolic test result: 0 passed; 1 failed; time: 0.49s
```

**To Reproduce**

Check out https://github.com/igorganich/damn-vulnerable-defi-halmos/blob/master/test/naive-receiver/NaiveReceiver.t.sol

Run `halmos --function test_naiveReceiver`

**Environment:**
 - OS: mac
 - Python version: Python 3.12.1
 - Halmos and other dependency versions: f7ff1f8323207334c35da1542005ae4b77d23ae3

0xkarmacoma:
fixes #338

### DIFF
--- a/src/halmos/constants.py
+++ b/src/halmos/constants.py
@@ -4,3 +4,7 @@ VERBOSITY_TRACE_PATHS = 4
 VERBOSITY_TRACE_CONSTRUCTOR = 5
 
 MAX_MEMORY_SIZE = 2**20
+
+# to be compatible with startHoax(address),
+# see https://github.com/a16z/halmos/issues/338
+MAX_ETH = 1 << 128
--- a/src/halmos/sevm.py
+++ b/src/halmos/sevm.py
@@ -62,7 +62,7 @@ from .calldata import FunctionInfo
 from .cheatcodes import Prank, halmos_cheat_code, hevm_cheat_code
 from .config import Config as HalmosConfig
 from .console import console
-from .constants import MAX_MEMORY_SIZE
+from .constants import MAX_ETH, MAX_MEMORY_SIZE
 from .exceptions import (
     AddressCollision,
     EvmException,
@@ -126,6 +126,7 @@ from .utils import (
     is_concrete,
     is_f_sha3_name,
     match_dynamic_array_overflow_condition,
+    render_address,
     restore_precomputed_hashes,
     sha3_inv,
     str_opcode,
@@ -1654,7 +1655,14 @@ class Exec:  # an execution path
         self.path.append(Select(EMPTY_BALANCE, addr) == ZERO)
 
         # practical assumption on the max balance per account
-        self.path.append(ULT(value, con(2**96)))
+        # (only for symbolic values, otherwise we may append false constraints)
+        if is_bv_value(value):
+            if (v := value.as_long()) > MAX_ETH:
+                warn(
+                    f"address {render_address(addr)} has balance {v} greater than {MAX_ETH=}"
+                )
+        else:
+            self.path.append(ULE(value, con(MAX_ETH)))
 
         return value
 
--- a/tests/expected/all.json
+++ b/tests/expected/all.json
@@ -885,6 +885,33 @@
                 "num_paths": null,
                 "time": null,
                 "num_bounded_loops": null
+            },
+            {
+                "name": "check_deal_over_max_eth_concrete()",
+                "exitcode": 1,
+                "num_models": 1,
+                "models": null,
+                "num_paths": null,
+                "time": null,
+                "num_bounded_loops": null
+            },
+            {
+                "name": "check_deal_over_max_eth_mixed(address)",
+                "exitcode": 0,
+                "num_models": 0,
+                "models": null,
+                "num_paths": null,
+                "time": null,
+                "num_bounded_loops": null
+            },
+            {
+                "name": "check_deal_under_max_eth_mixed(address)",
+                "exitcode": 1,
+                "num_models": 1,
+                "models": null,
+                "num_paths": null,
+                "time": null,
+                "num_bounded_loops": null
             }
         ],
         "test/Extcodehash.t.sol:ExtcodehashTest": [
--- a/tests/regression/test/Deal.t.sol
+++ b/tests/regression/test/Deal.t.sol
@@ -4,6 +4,7 @@ pragma solidity >=0.8.0 <0.9.0;
 import "forge-std/Test.sol";
 
 contract DealTest is Test {
+    uint256 constant MAX_ETH = 1 << 128;
     C c;
 
     function check_deal_1(address payable receiver, uint amount) public {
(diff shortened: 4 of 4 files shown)


0xkarmacoma:
- adds an explicit warning when we vm.deal an amount too large (could fail if we wanted to be more conservative)
- increases the tolerance in balance_of to be more aligned with `startHoax()`, but users can still run into the issue if they deal a large amount like `10 ** 75`
- only add the constraint on symbolic balances to avoid adding infeasible constraints (but as the test shows, it means some tests may pass which could be considered surprising)

src/halmos/sevm.py
```diff
@@ -1654,7 +1655,14 @@ def balance_of(self, addr: Word) -> Word:
         self.path.append(Select(EMPTY_BALANCE, addr) == ZERO)
 
         # practical assumption on the max balance per account
-        self.path.append(ULT(value, con(2**96)))
+        # (only for symbolic values, otherwise we may append false constraints)
+        if is_bv_value(value):
+            if (v := value.as_long()) > MAX_ETH:
+                warn(
```
0xkarmacoma, line 1661: after sleeping on it, I wonder if we should raise an exception here instead of printing a warning. Otherwise it's a bit strange to have a constraint like `ULE(value, con(MAX_ETH))`, while some values may in fact be higher than MAX_ETH

daejunpark: yeah we wouldn't want to continue once the value becomes larger than the max value, as any counterexamples wouldn't make sense anyway. so it might be better to raise an exception here, and let the outer loop handle it properly depending on context.

src/halmos/sevm.py
```diff
@@ -1654,7 +1655,14 @@ def balance_of(self, addr: Word) -> Word:
...
+                    f"address {render_address(addr)} has balance {v} greater than {MAX_ETH=}"
+                )
+        else:
+            self.path.append(ULE(value, con(MAX_ETH)))
```
daejunpark, line 1665: just note:  if we decide to raise in the true branch, it would be more consistent to check the path condition here and raise as well if it's infeasible. but it might incur too much overhead to call the smt solver everytime.

src/halmos/constants.py
```diff
@@ -4,3 +4,7 @@
 VERBOSITY_TRACE_CONSTRUCTOR = 5
 
 MAX_MEMORY_SIZE = 2**20
+
+# to be compatible with startHoax(address),
+# see https://github.com/a16z/halmos/issues/338
+MAX_ETH = 1 << 128
```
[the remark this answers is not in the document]
0xkarmacoma: each deal could be more than `1 << 128` if you allow it to be more, the limit has to be set somewhere. Alternatively we can make it configurable, but I don't think that's necessary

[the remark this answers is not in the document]
daejunpark: ah, i meant the deal from hoax().

hoax(x);
hoax(y);
x.send(y); // <-- we wouldn't want to see failure here.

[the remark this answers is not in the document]
0xkarmacoma: I understand that, but my point is that the limit has to be set somewhere, and it is easier to explain that hoax already deals MAX_ETH so you can't add to that. How about we deploy with the current MAX_ETH and raise it if it is still a concern? (and hopefully the tighter constraint helps in some cases)

src/halmos/sevm.py
0xkarmacoma: I see, `ULE(value, con(MAX_ETH))` may get simplified to `False` and then we'd still get the error message... 🫠

0xkarmacoma: can you check again with https://github.com/a16z/halmos/pull/501/commits/a6229542db6483a4de4d7f5da7133b457738c110 ?

daejunpark: looks good to me!

daejunpark (approved):

src/halmos/constants.py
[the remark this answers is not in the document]
daejunpark: i'm okay with that! (sorry for the delayed reply, i somehow thought that this pr had been already merged.)

PR #505, opened 2025-05-03
Issue365

Jayakumar2812:
### Added support for  below VM cheatcodes : 
*  ```function randomAddress() external returns (address);```
* ``` function randomBool() external view returns (bool);```
* ``` function randomBytes(uint256 len) external view returns (bytes memory);```
* ``` function randomBytes4() external view returns (bytes4);```
* ``` function randomBytes8() external view returns (bytes8);```
* ``` function randomInt() external view returns (int256);```
* ``` function randomInt(uint256 bits) external view returns (int256);```
* ``` function randomUint() external returns (uint256);```
* ``` function randomUint(uint256 min, uint256 max) external returns (uint256);```
* ``` function randomUint(uint256 bits) external view returns (uint256);```
* ``` function setArbitraryStorage(address target) external;```

### Test Cases : 
* Added test cases to verify the functionality of each cheat code.

### Desgin Choice:
* I created a separate function specifically for ``` create_bytes8 () ``` to handle the generation of an 8-byte array, instead of using a generic function for fixed-size byte arrays. This ensures better clarity and specialization for handling 8-byte arrays.Let me know if this needs to be refactored

### DIFF
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -18,7 +18,6 @@ from z3 import (
     Implies,
     Not,
     Or,
-    eq,
     is_bv,
     is_false,
     simplify,
@@ -56,6 +55,7 @@ from .utils import (
     extract_bytes,
     extract_funsig,
     extract_string_argument,
+    extract_word,
     f_ecrecover,
     green,
     hexify,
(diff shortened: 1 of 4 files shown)


src/halmos/cheatcodes.py
```diff
@@ -405,12 +408,18 @@ def create_uint(ex, arg, **kwargs):
...
+        name = "vmRandomUint"
     return ByteVec(uint256(create_generic(ex, bits, name, f"uint{bits}")))
 
 
 def create_uint256(ex, arg, **kwargs):
-    name = name_of(extract_string_argument(arg, 0))
+    try:
+        name = name_of(extract_string_argument(arg, 0))
+    except (IndexError, NotConcreteError):
+        name = "vmRandomUint256"
```
0xkarmacoma, line 422: would be better to avoid the try..catch pattern, maybe something with a default kwarg?

src/halmos/cheatcodes.py
```diff
@@ -461,15 +482,61 @@ def create_bytes32(ex, arg, **kwargs):
...
+def create_uint256_min_max(ex, arg, **kwargs):
+    symbolic_value = create_generic(ex, 256, "vmRandomUint_min_max", "uint256")
+
+    min_value = uint256(extract_bytes(arg, 4 + 32 * 1, 32))
+    max_value = uint256(extract_bytes(arg, 4 + 32 * 2, 32))
+
+    min_value = uint256(min_value).as_z3()
+    max_value = uint256(max_value).as_z3()
+
+    if eq(simplify(min_value), simplify(max_value)):
```
0xkarmacoma, line 509: this simplifies both but then discards the result, it would be better to save the simplified values

src/halmos/cheatcodes.py
```diff
@@ -461,15 +482,61 @@ def create_bytes32(ex, arg, **kwargs):
...
 
 
+def create_uint256_min_max(ex, arg, **kwargs):
+    symbolic_value = create_generic(ex, 256, "vmRandomUint_min_max", "uint256")
+
+    min_value = uint256(extract_bytes(arg, 4 + 32 * 1, 32))
+    max_value = uint256(extract_bytes(arg, 4 + 32 * 2, 32))
+
+    min_value = uint256(min_value).as_z3()
+    max_value = uint256(max_value).as_z3()
```
0xkarmacoma, line 507: no need for the extra `uint256` calls, they are already `uint256`

tests/regression/test/HalmosCheatCode.t.sol
```diff
@@ -84,6 +126,38 @@ contract HalmosCheatCodeTest is SymTest, Test {
...
+        uint256 r;
+        assembly {
+            r := returndatasize()
+        }
+        assert(r == 32);
+
+        uint256 x_uint = uint256(uint64(x));
+        assertLe(x_uint, type(uint64).max);
+        uint y; assembly { y := x }
+        assertLe(y, type(uint256).max);
```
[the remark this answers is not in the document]
0xkarmacoma: ok sorry I see that this is just following the existing pattern for consistency

0xkarmacoma (approved):
thank you, I addressed the minor comments directly!

PR #507, opened 2025-05-06
feat: improve handling storage slot hashes

daejunpark:
improves handling of keccak hashes and decoding storage slots

- now concrete hash values are returned whenever available.
- original hash expressions are recorded and later retrieved for decoding storage slots.
- more known/common hash values are considered, specifically slots for m[0] and m[1] for any m with slot(m) < 256.
- skips tracking hash values with large preimages (> 1024 bits), which are likely to be creation bytecode, to reduce overhead in smt query generation and solving.


src/halmos/cheatcodes.py
```diff
@@ -866,8 +866,8 @@ def handle(sevm, ex, arg: ByteVec, stack) -> ByteVec | None:
 
         # vm.sign(uint256 privateKey, bytes32 digest) returns (uint8 v, bytes32 r, bytes32 s)
         elif funsig == hevm_cheat_code.sign_sig:
-            key = extract_bytes(arg, 4, 32)
-            digest = extract_bytes(arg, 4 + 32, 32)
+            key = uint256(extract_bytes(arg, 4, 32)).as_z3()
+            digest = uint256(extract_bytes(arg, 4 + 32, 32)).as_z3()
```
daejunpark, line 870: q: any better way to do the same thing?

src/halmos/utils.py
```diff
@@ -940,283 +944,82 @@ class EVM:
...
+        return new_map
+
+
+def mk_precomputed_keccak_registry() -> OffsetMap:
+    m = OffsetMap()
+    for k, v in keccak256_256.items():
+        v = f_sha3_256(con(v))
+        m[k] = v
+    for k, (v1, v2) in keccak256_512.items():
+        v = f_sha3_512(con(v1 << 256 + v2, size_bits=512))
```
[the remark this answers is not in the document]
daejunpark: 👍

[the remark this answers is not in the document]
daejunpark: note: test passed because only the case with `v2 = 0` was tested.

src/halmos/cheatcodes.py
0xkarmacoma: It is readable but a bit too general. This will:
- grab the slice as a ByteVec
- unwrap() the slice, returning the underlying bytes or z3 object after calling simplify()
- wrap it in a BV (will call simplify() on it again if it is a z3 object)
- then finally wrap the underlying value in a z3 object if it is an int or return it

A more efficient way to do this would be something like:

```py
word = extract_word(arg, 4)
key = word if is_bv(word) else con(word)
```

or as a one-liner:

```py
key = (word := extract_word(arg, 4)) if is_bv(word) else con(word)
```

0xkarmacoma:
haven't reviewed yet, do you think this will help with this issue? https://github.com/a16z/halmos/issues/393

src/halmos/cheatcodes.py
daejunpark: yes i was afraid of that, and curious if there's a helper function for this. i will go back to using bytes_to_bv_value().

daejunpark: https://github.com/a16z/halmos/pull/507/commits/32c978a51a6866a2cb358975cb4ff6d9431783db

src/halmos/sevm.py
```diff
@@ -1387,6 +1387,82 @@ def digest(self) -> bytes:
...
+        return self._hashes.keys()
+
+    def values(self) -> Iterator[int]:
+        return self._hashes.values()
+
+    def copy(self) -> "KeccakRegistry":
+        new_registry = KeccakRegistry()
+        new_registry._hashes = self._hashes.copy()
+        new_registry._exprs = self._exprs.copy()
+        return new_registry
```
0xkarmacoma, line 1448: why would we need to copy this? seems to me this is an append-only structure so ideally copy(self) would return self

src/halmos/sevm.py
```diff
@@ -1387,6 +1387,82 @@ def digest(self) -> bytes:
...
+
+    def __init__(self):
+        self._hashes: dict[BitVecRef, int] = {}  # hash -> id
+        self._exprs = OffsetMap()  # hash -> expr
+
+    def __getitem__(self, key: BitVecRef) -> int:
+        return self._hashes[key]
+
+    def __setitem__(self, key: BitVecRef, value: int) -> None:
+        self._hashes[key] = value
```
0xkarmacoma, line 1427: would it make sense to add `assert (existing := self._hashes.get(key)) is None or existing == value`?

src/halmos/sevm.py
```diff
@@ -1387,6 +1387,82 @@ def digest(self) -> bytes:
...
+    def values(self) -> Iterator[int]:
+        return self._hashes.values()
+
+    def copy(self) -> "KeccakRegistry":
+        new_registry = KeccakRegistry()
+        new_registry._hashes = self._hashes.copy()
+        new_registry._exprs = self._exprs.copy()
+        return new_registry
+
+    def record(self, expr, hash_value) -> None:
```
0xkarmacoma, line 1450:
```suggestion
    def record(self, expr: BitVecRef, hash_value: bytes) -> None:
```

src/halmos/sevm.py
```diff
@@ -1387,6 +1387,82 @@ def digest(self) -> bytes:
...
+        hash_value = int.from_bytes(hash_value)
+        self._exprs[hash_value] = expr
+
+    def reverse_lookup(self, hash_value) -> BitVecRef:
```
0xkarmacoma, line 1454: can you add a type annotation here? Not immediately clear to me if hash_value can be int, bytes or both

src/halmos/utils.py
```diff
@@ -941,283 +949,75 @@ class EVM:
...
+
+        Args:
+            offset_bits: Number of bits to use for the offset part of the key.
+                        The remaining bits are used as the main dictionary key.
+        """
+        self._map = {}
+        self._offset_bits = offset_bits
+        self._mask = (1 << offset_bits) - 1  # Creates a mask of offset_bits 1's
+
+    def __getitem__(self, k: int) -> tuple[int, int]:
```
0xkarmacoma, line 978:
```suggestion
    def __getitem__(self, k: int) -> tuple[int | None, int | None]:
```

src/halmos/utils.py
```diff
@@ -941,283 +949,75 @@ class EVM:
...
+        Args:
+            k: The integer key to store
+            v: The value to store
+        """
+        self._map[k >> self._offset_bits] = (v, k & self._mask)
+
+    def copy(self) -> "OffsetMap":
+        new_map = OffsetMap(self._offset_bits)
+        new_map._map = self._map.copy()
+        return new_map
```
0xkarmacoma, line 1008: same as before, if this is expected to be append-only can we avoid copies?

src/halmos/sevm.py
daejunpark: um, no, we cannot share it across different paths, where there may exist different hash expressions that are evaluated to the same value.  we don't need deepcopy, but still need shallow copy.

src/halmos/sevm.py
daejunpark: yep i forgot about it

src/halmos/sevm.py
daejunpark: https://github.com/a16z/halmos/pull/507/commits/b80afffa27f7ce4edb24d60be9a6ae1cfa9d80cd

src/halmos/utils.py
daejunpark: this copy() method is used when creating a branch.  if we just return self, any new items added after the branch creation would appear in both paths, which would break the current code logic.

that said, we might be able to improve the logic to avoid copies, e.g., by annotating map items with their originating path info. let's discuss this optimization possibility separately.

0xkarmacoma:
> > haven't reviewed yet, do you think this will help with this issue? #393
> 
> good point! i think so. would you mind adding it as a test, if you already have a reproduce?

in [26624b0](https://github.com/a16z/halmos/pull/507/commits/26624b00a435498093805aa77c44528ad8567b3a)

(simplified as much as possible to avoid adding new dependencies)

src/halmos/sevm.py
```diff
@@ -1387,6 +1387,82 @@ def digest(self) -> bytes:
...
+    """
+
+    def __init__(self):
+        self._hashes: dict[BitVecRef, int] = {}  # hash -> id
+        self._exprs = OffsetMap()  # hash -> expr
+
+    def __getitem__(self, key: BitVecRef) -> int:
+        return self._hashes[key]
+
+    def __setitem__(self, key: BitVecRef, value: int) -> None:
```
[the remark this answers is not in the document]
daejunpark: https://github.com/a16z/halmos/pull/507/commits/1f6bbabccc30493769bedcc5df2b9dc2f1ab56b0

daejunpark:
re: https://github.com/a16z/halmos/pull/507#discussion_r2076548238

you have a good point, and we can develop the idea further.

note though, that there's another usage of the mapping:  hash distinction conditions are generated only when the given hash expression is new in the current path.  once we stop copying, a simple membership check won't work for that purpose.  that said, it might be possible to adjust the current logic accordingly.

let me merge this for now, and we can continue working on it.

PR #517, opened 2025-05-14
feat: add createCalldata with address

daejunpark:
adds an alternative version of createCalldata that takes a target address instead of a contract name.

for now, it supports only concrete addresses.

### DIFF
--- a/examples/simple/test/Reentrancy.t.sol
+++ b/examples/simple/test/Reentrancy.t.sol
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity >=0.8.0 <0.9.0;
+
+import "forge-std/Test.sol";
+import {SymTest} from "halmos-cheatcodes/SymTest.sol";
+
+import {ERC1155, IERC1155} from "openzeppelin/token/ERC1155/ERC1155.sol";
+import {ERC1155Holder} from "openzeppelin/token/ERC1155/utils/ERC1155Holder.sol";
+import {ReentrancyGuard} from "openzeppelin/utils/ReentrancyGuard.sol";
+
+uint256 constant TOKEN_ID = 1;
+
+contract ERC1155Mock is ERC1155 {
+    constructor() ERC1155("URI") {
+        _mint(msg.sender, TOKEN_ID, 1_000_000_000 ether, "");
+    }
+}
+
+contract ReentrancyVault is ReentrancyGuard, ERC1155Holder {
+    mapping(address => uint) balances;
+    IERC1155 token;
+
+    constructor (IERC1155 _token) {
+        token = _token;
+    }
+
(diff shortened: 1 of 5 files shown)


src/halmos/cheatcodes.py
```diff
@@ -289,12 +289,42 @@ def snapshot_state(
     return ByteVec(balance_hash + code_hash + storage_hash + path_hash)
 
 
-def create_calldata_contract(ex, arg, sevm, stack):
+def _get_contract_name(ex, arg) -> tuple[str | None, str | None]:
+    # TODO: support symbolic target using sevm.resolve_address_alias()
+    addr = con(
+        int_of(extract_bytes(arg, 4, 32), "symbolic address for SVM.createCalldata()"),
+        size_bits=160,
+    )
```
0xkarmacoma, line 297:
```suggestion
    addr = con_addr(
        int_of(arg.get_word(4), "symbolic address for SVM.createCalldata()"),
    )
```

src/halmos/cheatcodes.py
```diff
@@ -289,12 +289,42 @@ def snapshot_state(
...
+    code = ex.code[addr]
```
0xkarmacoma, line 298: this probably fails with a KeyError if the address is not in code? consider something like

```suggestion
    code = ex.code.get(addr)
    if code is None:
        raise HalmosException(f"createCalldata: no contract at address {hexify(addr)})
```

examples/simple/test/Reentrancy.t.sol
```diff
@@ -0,0 +1,112 @@
...
+
+        bytes memory data = svm.createCalldata(target);
+        target.call(data);
+
+        return this.onERC1155Received.selector;
+    }
+}
+
+/// @custom:halmos --early-exit
+contract ReentrancyTest is SymTest, Test, ERC1155Holder {
```
0xkarmacoma, line 71: nice test 👌

0xkarmacoma (approved):

src/halmos/cheatcodes.py
daejunpark: good catch!

src/halmos/cheatcodes.py
daejunpark: nice, let me open a separate pr that fixes this throughout the codebase.

src/halmos/cheatcodes.py
daejunpark: fixed: https://github.com/a16z/halmos/pull/517/commits/20b10edcff977be9e912947bca59f29e99a8282a

src/halmos/cheatcodes.py
daejunpark: fixed: https://github.com/a16z/halmos/pull/519

PR #518, opened 2025-05-14
chore: set up Trusted Publishing for publish-pypi.yml

0xkarmacoma:


mattg-a16z (approved):
Grants the workflow access to create an OIDC token, which then is used by the workflow to identify itself to PyPi. That should work correctly, both PyPi's docs and GH's docs (https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect) confirm this. 

LGTM

PR #519, opened 2025-05-16
refactor: extract_bytes(x, y, 32) => x.get_word(y)

daejunpark:
extract_bytes(x, y, 32) => x.get_word(y) when x is ByteVec

### DIFF
--- a/src/halmos/cheatcodes.py
+++ b/src/halmos/cheatcodes.py
@@ -90,21 +90,17 @@ def name_of(x: str) -> str:
     return re.sub(r"\s+", "_", x)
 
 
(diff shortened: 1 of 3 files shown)


src/halmos/cheatcodes.py
```diff
@@ -292,7 +288,7 @@ def snapshot_state(
 def _get_contract_name(ex, arg) -> tuple[str | None, str | None]:
     # TODO: support symbolic target using sevm.resolve_address_alias()
     addr = con(
```
0xkarmacoma, line 290: could be a con_addr while we're at it

0xkarmacoma (approved):

PR #521, opened 2025-05-17
feat: instruction coverage reporting

daejunpark:
This adds instruction coverage reporting, generating LCOV format reports to track which lines of code are executed during testing.

Each time an instruction is executed, its pc is used to retrieve the corresponding file path and line number, using the [source mapping info](https://docs.soliditylang.org/en/latest/internals/source_mappings.html). The resulting coverage report shows how many times each line is executed across all tests.

an html page can be generated from the lcov data, e.g.,:
```sh
genhtml coverage.info -o coverage
```


src/halmos/mapper.py
```diff
@@ -79,6 +82,112 @@ def __call__(cls, *args, **kwargs):
...
+        with open(filepath, "rb") as f:
+            while True:
+                chunk = f.read(8192)
+                if not chunk:
+                    break
+                for i, b in enumerate(chunk):
+                    if b == ord("\n"):
+                        newline_offsets.append(offset + i)
+                offset += len(chunk)
+        return newline_offsets
```
0xkarmacoma, line 188:
```
hyperfine 'python iter.py /usr/share/dict/words' 'python readlines.py /usr/share/dict/words' 'python chunks.py /usr/share/dict/words'
Benchmark 1: python iter.py /usr/share/dict/words
  Time (mean ± σ):      77.1 ms ±   2.2 ms    [User: 66.9 ms, System: 8.4 ms]
  Range (min … max):    74.2 ms …  84.4 ms    37 runs
 
Benchmark 2: python readlines.py /usr/share/dict/words
  Time (mean ± σ):      71.6 ms ±   1.8 ms    [User: 58.6 ms, System: 11.1 ms]
  Range (min … max):    69.0 ms …  77.8 ms    40 runs
 
Benchmark 3: python chunks.py /usr/share/dict/words
  Time (mean ± σ):     254.7 ms ±  14.7 ms    [User: 244.8 ms, System: 7.8 ms]
  Range (min … max):   238.3 ms … 290.3 ms    11 runs
 
Summary
  python readlines.py /usr/share/dict/words ran
    1.08 ± 0.04 times faster than python iter.py /usr/share/dict/words
    3.56 ± 0.22 times faster than python chunks.py /usr/share/dict/words
```

This is readlines.py:

```
import sys

newline_offsets = []
offset = 0

with open(sys.argv[1], "r") as f:
    lines = f.readlines()
    for line in lines:
        offset += len(line)
        newline_offsets.append(offset - 1)
```

This is iter.py:

```
import sys

newline_offsets = []
offset = 0

with open(sys.argv[1], "r") as f:
    for line in f:
        offset += len(line)
        newline_offsets.append(offset - 1)
```

This is chunks.py:

```
import sys

newline_offsets = []
offset = 0

with open(sys.argv[1], "rb") as f:
    while True:
        chunk = f.read(8192)
        if not chunk:
            break
        for i, b in enumerate(chunk):
            if b == ord("\n"):
                newline_offsets.append(offset + i)
        offset += len(chunk)
```

Good news is that the most straightforward version is the fastest! But we may not be handling CRLF line terminators on Windows properly

src/halmos/mapper.py
```diff
@@ -168,12 +277,13 @@ def create_build_out_map_code(self):
 
                 size = len(hexcode) // 2  # byte length
                 placeholders = self.get_placeholders(deployed)
-                code_data = (hexcode, placeholders, contract_name, filename)
+                source_map = deployed.get("sourceMap", "")
+                code_data = (hexcode, placeholders, contract_name, filename, source_map)
                 self._build_out_map_code[size].append(code_data)
 
-    def get_by_code(self, bytecode: ForwardRef("ByteVec")) -> tuple[str, str]:
+    def get_by_code(self, bytecode: ForwardRef("ByteVec")) -> tuple[str, str, str]:
```
0xkarmacoma, line 284:
```suggestion
    def get_by_code(self, bytecode: ForwardRef("ByteVec")) -> tuple[str | None, str | None, str | None]:
```

0xkarmacoma (approved):
remarkably straightforward! 👏 just left some minor comments, no blockers

src/halmos/mapper.py
```diff
@@ -79,6 +82,112 @@ def __call__(cls, *args, **kwargs):
...
+        """
+        if (line_offsets := self._line_offsets.get(filepath)) is None:
+            line_offsets = self._index_newlines(filepath)
+            self._line_offsets[filepath] = line_offsets
+
+        return bisect.bisect_right(line_offsets, byte_offset) + 1
+
+    def get_location(
+        self, file_id: int, byte_offset: int
+    ) -> tuple[str | None, int | None]:
```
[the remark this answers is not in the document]
daejunpark: now line number can be None

src/halmos/mapper.py
daejunpark: thanks for this! i went with the iter.py approach to reduce memory footprint for large files.

i also fixed some subtle indexing issues and added unit tests: [1650686](https://github.com/a16z/halmos/pull/521/commits/1650686df6f5a28f2975a245c74a93d512561bc7)
(vibe coding isn't yet ready to handle this well.)

PR #527, opened 2025-06-02
fix: too many open files

### LINKED ISSUE #523: fix: segfault related to too many open files
@mattg-a16z reported that a long-running test (about a week) failed with segfault related to too many open files.  i suspect it's probably because all smt query files are generated in a single directory.  a nested directory structure would be needed to store a large number of smt files.

daejunpark:
fixes #523 

the error "OSError: [Errno 24] Too many open files" occurred when many solver processes were killed due to timeout, where the file descriptors of python subprocess objects weren't properly closed, leading to resource leaks.

note that when a process is killed via psutil, the file descriptors may remain open on the Python side, e.g.,:
```python
p = Popen(['ls'], stdout=PIPE, stderr=PIPE)                                                          
psutil.Process(p.pid).kill()
print(p.stdout.closed)  # False - the pipe is still open!
```

### DIFF
--- a/src/halmos/processes.py
+++ b/src/halmos/processes.py
@@ -68,10 +68,13 @@ class PopenFuture(concurrent.futures.Future):
                 self.returncode = self.process.returncode
             except subprocess.TimeoutExpired as e:
                 self._exception = e
-                self.cancel()
             except Exception as e:
                 self._exception = e
             finally:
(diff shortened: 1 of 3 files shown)


daejunpark:
> great catch!

credit: claude-4-sonnet-thinking

PR #531, opened 2025-06-04
implement explicit config source precedence

0xkarmacoma:
fixes the annoying "can not provide both --solver-command and --solver" error


0xkarmacoma:
fixes the snekmate tests: https://github.com/a16z/halmos/actions/runs/15430195860/job/43426353850

daejunpark (approved):

Issue #514, opened 2025-05-12
fix: exclude all special functions from `targetContract(address(this))`


0xkarmacoma:
See context in https://github.com/a16z/halmos/issues/506

I propose that we keep `invariant_` functions stateless (i.e. they are never considered state-changing target functions that can be part of a call sequence). Other functions in the test contract are fine to include.

We could also consider automatically including `handler_` functions -- but maybe that's an unwelcome departure from the foundry convention?

0xkarmacoma:
in [#535](https://github.com/a16z/halmos/pull/535)

Issue #506, opened 2025-05-05
Allow `targetContract(address(this));` to perform Invariant Testing on Scaffolded Contracts


GalloDaSballo:
**Is your feature request related to a problem? Please describe.**
Per this merged foundry feature: https://github.com/foundry-rs/foundry/issues/10253

We expect `targetContract(address(this));` to cause the fuzzer to use every function exposed by this as a handler

And we expect all invariants to be specified on `this`

**Describe the solution you'd like**

An extremely common way of writing invariant tests uses Target Function, explicit functions that tell the fuzzer how to explore state

Source:
https://github.com/Recon-Fuzz/create-chimera-app (Recon)
https://github.com/perimetersec/arachne (PerimeterSec)
https://github.com/euler-xyz/euler-vault-kit/tree/master/src (Enigma Dark)
https://github.com/CodeHawks-Contests/2025-02-gamma/blob/main/test/fuzzing/echidna/CryticToFoundry.sol (Guardian Audits)

As it stands targetFunctions doesn't allow passing address(this)

This prevents these testing suites from being easily reusable with Halmos

**Describe alternatives you've considered**
This is an additional feature that makes Halmos competitive with Foundry, Echidna and Medusa

I'm unaware of any alternative that makes Halmos as powerful

**Additional context**
You can attempt running all tools on: https://github.com/Recon-Fuzz/create-chimera-app

Which is a template that is compatible with all of them
However, Halmos Invariant Testing will fail due to a lack of compatibility with `targetContract(address(this));`

0xkarmacoma:
Testing with:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.17;

import "forge-std/Test.sol";

contract InvariantTest is Test {
    bool fooCalled;

    function setUp() public {
        targetContract(address(this));
    }

    function foo() public {
        fooCalled = true;
    }

    function invariant_this() public {
        assertFalse(fooCalled);
    }
}
```

On foundry 1.0.0-rc:

```
Ran 1 test for test/Invariant.t.sol:FuzzBytes4
[FAIL: failed to set up invariant testing environment: No contracts to fuzz.] invariant_this() (runs: 0, calls: 0, reverts: 0)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 14.84ms (3.71ms CPU time)
```

But on foundry 1.1.0-stable:

```
Ran 1 test for test/Invariant.t.sol:FuzzBytes4
[...]

PR #535, opened 2025-06-10
allow targetContract(address(this)), exclude special functions automa…

0xkarmacoma:
…tically

https://github.com/a16z/halmos/issues/506
https://github.com/a16z/halmos/issues/514

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -1343,7 +1343,7 @@ def resolve_target_contracts(ctx: InvariantTestingContext, ex: Exec) -> set[Addr
     # or targetContract(FOUNDRY_TEST) is provided.
     resolved_target_contracts = (
         resolved_target_contracts
-        if (target_selectors.get(FOUNDRY_TEST) or FOUNDRY_TEST in target_contracts)
+        if (FOUNDRY_TEST in target_contracts or target_selectors.get(FOUNDRY_TEST))
         else resolved_target_contracts - {FOUNDRY_TEST}
     )
 


tests/regression/test/InvariantTargetThis.t.sol
```diff
@@ -0,0 +1,52 @@
...
+        assertEq(check_foo_called, false);
+        assertEq(setUpSymbolic_called, false);
+        assertEq(test_foo_called, false);
+        assertEq(invariant_this_called, false);
+    }
+
+    // we expect a counterexample, showing that foo is indeed called
+    function invariant_targets_foo_included() public view {
+        assertEq(foo_called, false);
+    }
```
0xkarmacoma, line 51: some tests fail because we get either 1 or 2 counterexample (depth 1 vs depth 2). Any idea why? I suspect it has to do with visited states, at depth 2 we write the same value back to storage

daejunpark: you're right. in the generic storage mode, any storage updates are considered generating a new state. a workaround is to use --early-exit or --invariant-depth 1 for this test.

src/halmos/__main__.py
```diff
@@ -1340,22 +1339,18 @@ def resolve_target_contracts(ctx: InvariantTestingContext, ex: Exec) -> set[Addr
...
     resolved_target_contracts |= target_selectors.keys()
 
-    # Note: FOUNDRY_TEST is excluded unless a targetSelector() is specified for it, even if targetContract(FOUNDRY_TEST) is provided.
-    result = (
+    # Note: FOUNDRY_TEST is excluded unless a targetSelector() is specified for it
+    # or targetContract(FOUNDRY_TEST) is provided.
+    resolved_target_contracts = (
         resolved_target_contracts
-        if target_selectors.get(FOUNDRY_TEST)
+        if (target_selectors.get(FOUNDRY_TEST) or FOUNDRY_TEST in target_contracts)
```
daejunpark, line 1346: this should be faster in most cases, as it's uncommon to specify selectors without including in the target:
```suggestion
        if (FOUNDRY_TEST in target_contracts or target_selectors.get(FOUNDRY_TEST))
```

daejunpark (approved):

PR #526, opened 2025-06-02
Enhancement#54

Jayakumar2812:
## ✅ Enhancement: Add Full Support for `env`, `envOr`, and Array-Based Cheatcodes

This PR introduces complete support for all environment-based cheatcodes (`env*`, `envOr*`, and their array variants), fully aligned with Foundry’s specification.

---

### 🔧 What’s Added

- **Primitive Cheatcodes**:
  - `envInt(string)`
  - `envBytes32(string)`
  - `envAddress(string)`
  - `envBool(string)`
  - `envUint(string)`
  - `envString(string)`
  - `envBytes(string)`

- **Array Variants**:
  - `envInt(string,string)`
  - `envAddress(string,string)`
  - `envBool(string,string)`
  - `envBytes32(string,string)`
  - `envString(string,string)`
  - `envUint(string,string)`
  - `envBytes(string,string)`

- **`envOr` Variants**:
  - `envOr(string, address)`
  - `envOr(string, bool)`
  - `envOr(string, bytes)`
  - `envOr(string, bytes32)`
  - `envOr(string, int256)`
  - `envOr(string, uint256)`
  - `envOr(string, string)`
  - `envOr(string, string, address[])`
  - `envOr(string, string, bool[])`
  - `envOr(string, string, bytes[])`
  - `envOr(string, string, bytes32[])`
  - `envOr(string, string, int256[])`
  - `envOr(string, string, string[])`
  - `envOr(string, string, uint256[])`

- **Existence Check**:
  - `envExists(string)`

---

### 🛠️ Fixes & Changes

- **Corrected Misleading Signatures**:
  - In the previous [pr link](https://github.com/a16z/halmos/commit/081ea7402742ff852119184e80d94ebf1a1d9e6a) I added 
  ```solidity
    0x953C097E: "envBytes(bytes)",
    0x6C42F03F: "envBytes(bytes,bytes)",
  ```
  - The Foundry Book documented the following signatures incorrectly as using `bytes`:
    ```
    function envBytes(bytes calldata key)
    function envBytes(bytes calldata key, bytes calldata delimiter)
    ```
  - But the actual implementation uses `string` parameters:
    [Foundry source → cheatcodes.json](https://github.com/foundry-rs/foundry/blob/b42d5121ba3f03099b1d649ab71f2100a212d3f3/crates/cheatcodes/assets/cheatcodes.json#L4444)

  - Updated function selectors:
    ```solidity
    0x4D7BAF06: "envBytes(string)"
[...]

PR #536, opened 2025-06-10
chore: cleanup for #526

daejunpark:
some cleanups following up on #526:
- renamed `create_inv_*` to `inv_*`, as the `create_` prefix was misleading. (they only read env vars, without creating anything.)
- updated to consistently use absolute import paths.
- added missing license headers.


0xkarmacoma:
changes look good, ty!

PR #552, opened 2025-06-20
feat: save timeout queries for offline solving

daejunpark:
saves queries when solvers return unknown due to timeout, so that they can be solved offline.

also saves their call sequences and traces in separate files to provide context.

Screenshot 2025-06-20 at 1 11 53 PM

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -22,7 +22,6 @@ from datetime import timedelta
 from enum import Enum
 from functools import partial
 from importlib import metadata
-from tempfile import TemporaryDirectory
 from types import MappingProxyType
 
 from rich.console import Group
@@ -109,6 +108,7 @@ from halmos.solve import (
     InvariantTestingContext,
     PathContext,
     SolverOutput,
+    dirname,
     solve_end_to_end,
     solve_low_level,
 )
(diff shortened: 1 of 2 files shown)


src/halmos/__main__.py
```diff
@@ -923,6 +925,33 @@ def _solve_end_to_end_callback(
...
+            # save call sequence to a separate file for debugging context
+            if call_sequence := ctx.call_sequences.get(path_id):
+                call_seq_file = f"{timeout_query_file}.callseq"
+                try:
+                    with open(call_seq_file, "w") as f:
+                        f.write(call_sequence)
+                        if args.verbose >= VERBOSITY_TRACE_COUNTEREXAMPLE:
+                            f.write(f"\nTrace:\n{ctx.traces[path_id]}")
+                except Exception as e:
+                    print(f"Could not save call sequence: {e}")
```
0xkarmacoma, line 951: maybe `error` instead of print here too?

src/halmos/__main__.py
```diff
@@ -923,6 +925,33 @@ def _solve_end_to_end_callback(
             )
             return
 
+        path_id = solver_output.path_id
+
+        # handle "unknown" (timeout) result
+        if result == unknown:
+            # copy timeout query to separate timeout directory
+            timeout_dir = f"{dirname(ctx.solving_ctx.dump_dir)}-timeout"
```
0xkarmacoma, line 933: nice I was going to ask if this would play well with dump_dir 👌

0xkarmacoma (approved):

src/halmos/__main__.py
```diff
@@ -1172,6 +1200,11 @@ def run_test(ctx: FunctionContext) -> TestResult:
         f"bounds: [{', '.join([str(x) for x in dyn_params])}])"
     )
 
+    if exitcode == Exitcode.TIMEOUT.value:
+        # print timeout query directory information
+        timeout_dir = f"{dirname(ctx.solving_ctx.dump_dir)}-timeout"
+        print(f"Timeout queries saved in: {timeout_dir}")
```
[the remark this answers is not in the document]
daejunpark: good point, but `info` is enabled at a higher logger level than the default. i updated it to print with color_info.

PR #567, opened 2025-07-26
handle failed solver processes

daejunpark:
when an exception is raised in solver subprocesses, their result may not be added to the solver_outputs list. in the current implementation, if all processes fail due to exceptions, it may incorrectly return pass because no error solver outputs are recorded.

this pr adds logic to detect such cases by comparing the number of solver outputs with the number of solver processes. the detection is designed to distinguish these from early exit scenarios.

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -1193,6 +1193,11 @@ def run_test(ctx: FunctionContext) -> TestResult:
             REVERT_ALL,
             f"{funsig}: all paths have been reverted; the setup state or inputs may have been too restrictive.",
         )
+    elif len(ctx.solver_outputs) < len(submitted_futures):
+        # some solver results are missing, likely due to exceptions.
+        # not an early exit, as counter["sat"] > 0 would have been handled earlier.
+        passfail = color_error("[ERROR]")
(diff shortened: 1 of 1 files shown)


src/halmos/__main__.py
```diff
@@ -1193,6 +1193,11 @@ def run_test(ctx: FunctionContext) -> TestResult:
             REVERT_ALL,
             f"{funsig}: all paths have been reverted; the setup state or inputs may have been too restrictive.",
         )
+    elif len(ctx.solver_outputs) < len(submitted_futures):
```
0xkarmacoma, line 1196: I'm a bit concerned about relying on the number of solver outputs to infer that an error occured

> when an exception is raised in solver subprocesses, their result may not be added to the solver_outputs list

this seems to be the real issue? ideally we want to capture the errors and add failed outputs to `ctx.solver_outputs`, no?

daejunpark: > ideally we want to capture the errors and add failed outputs to ctx.solver_outputs, no?

yes, but it needs some more work, as the solver output object currently needs more info that isn't available at some places exceptions occur.

daejunpark:
superseded by #570

PR #570, opened 2025-07-30
create solver_output in case of errors too

0xkarmacoma:
surprisingly hard to figure out what the corresponding path_id/query was when we get an error, so for now let's make this optional in solver_output

this properly returns [ERROR] on tests that have solver_outputs with errors (could be PASS previously)

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -777,6 +777,16 @@ def run_message(
                 reset(solver)
 
 
+def is_benign_solving_error(e: Exception) -> bool:
+    match e:
+        case ShutdownError():
+            return True
+        case OSError(errno=9):  # BAD_FILE_DESCRIPTOR
+            return True
+        case _:
+            return False
+
+
 @dataclass(frozen=True)
 class CounterexampleHandler:
     """Handles potential assertion violations and generates counterexamples."""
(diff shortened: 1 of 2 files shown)


src/halmos/__main__.py
```diff
@@ -878,52 +910,19 @@ def _solve_end_to_end_callback(
...
-        # keep track of the solver outputs, so that we can display PASS/FAIL/TIMEOUT/ERROR later
-        ctx.solver_outputs.append(solver_output)
-
         if result == unsat:
             if solver_output.unsat_core:
                 ctx.append_unsat_core(solver_output.unsat_core)
             return
 
         if result == "err":
+            error(f"{solver_output.error=} ({solver_output.returncode=})")
```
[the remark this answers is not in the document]
0xkarmacoma: nah

src/halmos/__main__.py
```diff
@@ -856,13 +866,35 @@ def handle_assertion_violation(
...
+                error(f"encountered exception during assertion solving: {e!r}")
+            return SolverOutput.from_error(e, path_id=path_id, query_file=query_file)
+
+        try:
+            return future.result()
+        except Exception as e:
+            if not is_benign_solving_error(e):
+                error(f"encountered exception during assertion solving: {e!r}")
+            return SolverOutput.from_error(e, path_id=path_id, query_file=query_file)
+
```
daejunpark, line 895: originally, other OSErrors were re-raised from the catch block, but now they aren't.  does this not alter behavior because exceptions raised from callbacks are silently ignored anyway?

daejunpark (approved):

src/halmos/__main__.py
0xkarmacoma: well I'm not sure why we re-raised, but it seems to me this is a good place to capture errors and keep track of errors

PR #576, opened 2025-09-10
Enhance `div_xy_y` to support signed and unsigned division simplification

sidarth16:
### Summary
This PR updates the `div_xy_y` helper to handle simplifications for both unsigned (`DIV`) and signed (`SDIV`) EVM operations. It allows expressions like `(x * y) / x` or `(x * y) / y` to be simplified automatically, improving symbolic reasoning and reducing complexity in arithmetic analysis.

### Changes
- Added `signed: bool` parameter to `div_xy_y`.
- Implemented simplification logic for SDIV (signed division).
- Preserved unsigned simplification logic with overflow checks.
- Updated `arith` to call `div_xy_y` for both DIV and SDIV before falling back to generic operations.

### Notes
- This is an **internal optimization**; outputs remain identical to previous behavior.
- Existing unit tests cover correctness; no new tests are added in this PR.


.hypothesis/unicode_data/13.0.0/charmap.json.gz
0xkarmacoma, line 1: please don't add this file

tests/regression/foundry.lock
0xkarmacoma, line 1: not sure if foundry just started adding these, but let's keep this out of this PR please

0xkarmacoma (changes requested):
thanks @sidarth16, note that it would cool to add a couple unit tests to verify the functionality

deferring to @daejunpark for the actual div_xy_y logic safety.

PR #578, opened 2025-09-12
fix: symbolic storage variables in counterexample + add path id to co…

### LINKED ISSUE #577: symbolic storage variables are not included in the counterexample model
```solidity
pragma solidity ^0.8.15;

import "forge-std/Test.sol";

contract Counter {
    uint256 public number;

    function increment() public {
        number++;
    }
}

contract CounterHalmosTest is Test, SymTest {
    Counter public counter;

    function setUp () public {
        counter = new Counter();
        svm.enableSymbolicStorage(address(counter));
    }


    function check_isIncrementFeasible() public {
        vm.assume(counter.number() != 2);
        counter.increment();
        assertTrue(!(counter.number() == 2), "number is 2");
    }
}
```

Running halmos on this function returns:

```
Counterexample: ∅
[FAIL] check_isIncrementFeasible() (paths: 3, time: 0.14s, bounds: [])
```

Ideally we would see the `number` value that results in the counterexample.

## Investigation

We can see what's happening by inspecting the smt output, for instance:

```sh
halmos --function check_isIncrementFeasible --dump-smt-queries --solver z3 -vv        

# <snip>

Executing check_isIncrementFeasible
Found potential path with path_id=2 (fail flag set) VmAssertion(cond=Not(storage_0x00000000000000000000000000000000aaaa0002_256_00[0] == 1), msg='number is 2')
Checking path condition path_id=2
# of potential paths involving assertion violations: 1 / 3 (--solver-threads 20)
  Checking with external solver process
    /usr/local/bin/z3 /var/folders/p9/s6g8495n6dqdjq3nn8hkdn840000gn/T/check_isIncrementFeasible-r3_713gv/2.smt2 > /var/folders/p9/s6g8495n6dqdjq3nn8hkdn840000gn/T/check_isIncrementFeasible-r3_713gv/2.smt2.out
```

```sh
cat /var/folders/p9/s6g8495n6dqdjq3nn8hkdn840000gn/T/check_isIncrementFeasible-r3_713gv/2.smt2.out     
sat
(
  (define-fun storage_0x00000000000000000000000000000000aaaa0002_256_00 () (Array (_ BitVec 256) (_ BitVec 256))
    ((as const (Array (_ BitVec 256) (_ BitVec 256)))
  #x0000000000000000000000000000000000000000000000000000000000000001))
  (define-fun storage_0x7fa9385be102ac3eac297483dd6233d62b3e1496_256_00 () (Array (_ BitVec 256) (_ BitVec 256))
    ((as const (Array (_ BitVec 256) (_ BitVec 256)))
  #x0000000000000000000000000000000000000000000000000000000000000000))
  (define-fun storage_0x7fa9385be102ac3eac297483dd6233d62b3e1496_256_da8f873_01 () (Array (_ BitVec 256) (_ BitVec 256))
    (store ((as const (Array (_ BitVec 256) (_ BitVec 256)))
         #x0000000000000000000000000000000000000000000000000000000000000000)
       #x000000000000000000000000000000000000000000000000000000000000000c
       #x0000000000000000000000000000000000000000000000000000000000000001))
  (define-fun storage_0x00000000000000000000000000000000aaaa0002_256_2940072_04 () (Array (_ BitVec 256) (_ BitVec 256))
    (store ((as const (Array (_ BitVec 256) (_ BitVec 256)))
         #x0000000000000000000000000000000000000000000000000000000000000001)
       #x0000000000000000000000000000000000000000000000000000000000000000
       #x0000000000000000000000000000000000000000000000000000000000000002))
  (define-fun storage_0x7fa9385be102ac3eac297483dd6233d62b3e1496_256_af2032d_03 () (Array (_ BitVec 256) (_ BitVec 256))
    (store (store ((as const (Array (_ BitVec 256) (_ BitVec 256)))
                #x0000000000000000000000000000000000000000000000000000000000000000)
              #x000000000000000000000000000000000000000000000000000000000000000c
              #x0000000000000000000000000000000000000000000000000000000000000001)
       #x000000000000000000000000000000000000000000000000000000000000001f
       #x000000000000000000000000000000000000000000000000000000aaaa000201))
  (define-fun balance_00 () (Array (_ BitVec 160) (_ BitVec 256))
    ((as const (Array (_ BitVec 160) (_ BitVec 256)))
  #x0000000000000000000000000000000000000000000000000000000000000000))
  (define-fun balance_99ac464_01 () (Array (_ BitVec 160) (_ BitVec 256))
    (store ((as const (Array (_ BitVec 160) (_ BitVec 256)))
         #x0000000000000000000000000000000000000000000000000000000000000000)
       #x7fa9385be102ac3eac297483dd6233d62b3e1496
       #x0000000000000000000000000000000000000000ffffffffffffffffffffffff))
  (define-fun storage_0x7fa9385be102ac3eac297483dd6233d62b3e1496_256_1bfa2a2_02 () (Array (_ BitVec 256) (_ BitVec 256))
    (store (store ((as const (Array (_ BitVec 256) (_ BitVec 256)))
                #x0000000000000000000000000000000000000000000000000000000000000000)
              #x000000000000000000000000000000000000000000000000000000000000000c
              #x0000000000000000000000000000000000000000000000000000000000000001)
       #x000000000000000000000000000000000000000000000000000000000000001f
       #x0000000000000000000000000000000000000000000000000000000000000001))
)
```

The important bit is that `storage_0x00000000000000000000000000000000aaaa0002_256_00` has value `#x0000000000000000000000000000000000000000000000000000000000000001`, but it's not captured by the `halmos_var_pattern` regex in `solve.py`. 

Also note that there is plenty of other stuff in the output that would likely just add noise to the model.

0kn0t:
fixes #577 (only doesnt work for yices)
also includes path id in counterexample (for debug or advanced use cases)
Screenshot 2025-09-12 at 11 56 12 PM

### DIFF
--- a/src/halmos/__main__.py
+++ b/src/halmos/__main__.py
@@ -949,7 +949,7 @@ class CounterexampleHandler:
             print(ctx.traces[path_id], end="")
 
         if model.is_valid:
-            print(color_error(f"Counterexample: {model}"))
+            print(color_error(f"Counterexample: [{path_id}] {model}"))
             ctx.valid_counterexamples.append(model)
 
             # add the stacks from the temporary flamegraph to the global one
--- a/src/halmos/solve.py
+++ b/src/halmos/solve.py
@@ -61,7 +61,7 @@ ModelVariables = dict[str, ModelVariable]
 halmos_var_pattern = re.compile(
     r"""
     \(\s*define-fun\s+               # Match "(define-fun"
-    \|?((?:halmos_|p_)[^ |]+)\|?\s+  # Capture either halmos_* or p_*, optionally wrapped in "|"
+    \|?((?:halmos_|p_)[^ |]+|storage_[^ |]+_00)\|?\s+  # Capture halmos_* or p_* or storage_*_00 optionally wrapped in "|"
     \(\)\s+\(_\s+([^ ]+)\s+          # Capture the SMTLIB type (e.g., "BitVec 256")
     (\d+)\)\s+                       # Capture the bit-width or type argument
     (                                # Group for the value


src/halmos/__main__.py
```diff
@@ -949,7 +949,7 @@ def _solve_end_to_end_callback(
             print(ctx.traces[path_id], end="")
 
         if model.is_valid:
-            print(color_error(f"Counterexample: {model}"))
+            print(color_error(f"Counterexample: [{path_id}] {model}"))
```
daejunpark, line 952: this prints as `Counterexample: [N]`, but it might not be obvious to users what the number N means. could you make it to explicitly say that N is the path id?

src/halmos/solve.py
```diff
@@ -61,7 +61,7 @@ class ModelVariable:
 halmos_var_pattern = re.compile(
     r"""
     \(\s*define-fun\s+               # Match "(define-fun"
-    \|?((?:halmos_|p_)[^ |]+)\|?\s+  # Capture either halmos_* or p_*, optionally wrapped in "|"
+    \|?((?:halmos_|p_)[^ |]+|storage_[^ |]+_00)\|?\s+  # Capture halmos_* or p_* or storage_*_00 optionally wrapped in "|"
```
daejunpark, line 64: could you make including storage variables optional behind the --print-full-model flag?

for context: originally, storage variables used to be printed only when --print-full-model was enabled. that feature was somehow dropped in #437 (see the diff in `__main__.py`), and hasn't been brought back yet.

daejunpark: also, it looks like the current regex doesn't cover storage _mapping_ variables. could you confirm that as well?

daejunpark (commented):
thanks for the contribution, @0kn0t!  i've left a few comments.

PR #583, opened 2026-05-18
Support of OPCODE BLOBBASEFEE , BLOBHASH and CLZ

HugoDowsers:
This PR adds symbolic execution support for three EVM opcodes that were previously unhandled by Halmos, causing verification failures on contracts that use them.


daejunpark:
Thank you so much for this! In the meantime, could you add some tests for these new opcode instructions, similar to the existing ones here:
https://github.com/a16z/halmos/tree/main/tests/regression/test

HugoDowsers:
I've added some regression tests about the new opcodes. I needed to change the evm version in the foundry.toml

